PR #11, opened 2021-09-15
Use DssExecLib code v0.0.7

brianmcmichael:
Uses DssExecLib code tagged to `v0.0.7` release.

I've deployed a new DssExecLib to Goerli because the one that's existing here is pointing to `master`, which is a little different than the `0.0.7` release on mainnet. We should use this code because if there is any issue with using this version we will discover it here before we roll over to prod.


naszam (approved):
Checked:

- DssExecLib Goerli v0.0.7 (verified, [matches mainnet etherscan code](https://www.diffchecker.com/dPL0bGTw), 200 runs, licence AGPLv3)
- Code Changes Look Good

LGTM :+1:

julienmartinlevrai (approved):
LGTM, deployed code matches 0.0.7

PR #16, opened 2021-09-23
Add Gelato test dss-vest amounts

hexonaut:


src/Goerli-DssSpell.sol
```diff
@@ -19,13 +19,23 @@ pragma solidity 0.6.12;
 import "dss-exec-lib/DssExec.sol";
 import "dss-exec-lib/DssAction.sol";
 
+interface DssVestLike {
```
naszam, line 22: I think we don't need the DssVestLike interface as it's already included in [dss-interfaces](https://github.com/makerdao/dss-interfaces/blob/master/src/dss/VestAbstract.sol)

src/Goerli-DssSpell.t.sol
```diff
@@ -42,6 +42,22 @@ interface PsmAbstract {
     function buyGem(address usr, uint256 gemAmt) external;
 }
 
+interface DssVestLike {
```
naszam, line 45: Same as above.

src/Goerli-DssSpell.sol
hexonaut: That interface is out of date. `init` is now `create`, `usr()` is also missing. Just going to stick with the inline interfaces for now.

naszam: Oh! Good catch! 
We should fix that.

src/Goerli-DssSpell.sol
```diff
@@ -19,13 +19,23 @@ pragma solidity 0.6.12;
...
+    function create(address, uint256, uint256, uint256, uint256, address) external returns (uint256);
+    function file(bytes32, uint256) external;
```
naszam, line 24: I think we can remove it as not used in the spell.

src/Goerli-DssSpell.t.sol
```diff
@@ -42,6 +42,22 @@ interface PsmAbstract {
...
+    function fin(uint256) external view returns (uint256);
+    function mgr(uint256) external view returns (address);
+    function res(uint256) external view returns (uint256);
+    function tot(uint256) external view returns (uint256);
+    function rxd(uint256) external view returns (uint256);
+    function unrestrict(uint256) external;
+    function vest(uint256) external;
+}
+
+
```
naszam, line 60: nit: extra line

naszam (approved):
LGTM :+1:

PR #23, opened 2021-11-09
migrate all setUp, helper functions and interfaces to a base contract

iamchrissmith:
# Description

Move spell setup and help functions to a base contract and inherit it. 

# Contribution Checklist

- [ ] PR title starts with `(PE-<TICKET_NUMBER>)`
- [ ] Code approved
- [ ] Tests approved
- [ ] CI Tests pass

# Checklist

- [ ] Every contract variable/method declared as public/external private/internal
- [ ] Consider if this PR needs the `officeHours` modifier
- [ ] Verify expiration (`4 days + 2 hours` monthly and `30 days` for the rest)
- [ ] Verify hash in the description matches [here](https://emn178.github.io/online-tools/keccak_256.html)
- [ ] Validate all addresses used are in Goerli changelog or known
- [ ] Notify any external teams affected by the spell so they have the opportunity to review
- [ ] Deploy spell to Goerli `ETH_GAS="XXX" ETH_GAS_PRICE="YYY" make deploy`
- [ ] Ensure contract is verified on `Goerli` etherscan
- [ ] Change test to use Goerli spell address and deploy timestamp
- [ ] Keep `Goerli-DssSpell.sol` and `Goerli-DssSpell.t.sol` the same, but make a copy in `archive`
- [ ] `squash and merge` this PR

PR #25, opened 2021-11-10
Archive spells

iamchrissmith:
Make a script to archive the spell each week and organize the archive directory to have subdirectories that contain each week's archive

Tests should pass with: HardHat Fork pinned at block n. 5749312

currently dependent on https://github.com/makerdao/spells-goerli/pull/23, once that is merged, switch base to master


Makefile
```diff
@@ -5,3 +5,4 @@ clean  :; dapp clean
 test   :; ./test-dssspell.sh $(match)
 deploy :; make && dapp create DssSpell | xargs ./verify.py DssSpell
 flatten :; hevm flatten --source-file "src/Goerli-DssSpell.sol" > out/flat.sol
+archive-spell :; ./archive-spell.sh $(name)
```
naszam, line 8: I would name it just archive, to keep it short.

archive-spell.sh
```diff
@@ -0,0 +1,13 @@
...
+
+if [[ -z "$1" ]]; then
+  echo "You must provide a name option to name the directory"
+else
+  mkdir "./archive/$name"
+  cp "./src/Goerli-DssSpell.sol" "./archive/$name"
+  cp "./src/Goerli-DssSpell.t.sol" "./archive/$name"
+  cp "./src/Goerli-DssSpell.t.base.sol" "./archive/$name"
+  echo "Spell, tests and base copied to archive directory"
+fi
```
naszam, line 13: nit: missing new line

Makefile
iamchrissmith: i tried that and got `make: `archive' is up to date.` returned which made me think that `make archive` is a reserved command.

naszam: Oh! right.

naszam, line 8: nit: missing newline

naszam, line 8: should we keep the `dssspell` naming convention in the script file?

iamchrissmith: yeah i'll update the file name, good call

gbalabasquer: Done in my PR after merging this.

PR #29, opened 2021-11-15
PE-751: Weekly Executive 2021-11-19 (goerli)

naszam:
This reverts commit 2aefaed0f61f5004453190fce060adc15eeccdc3.
Adding commit back in on correct branch.

# Description

# Contribution Checklist

- [x] PR title starts with `(PE-<TICKET_NUMBER>)`
- [ ] Code approved
- [ ] Tests approved
- [ ] CI Tests pass

# Checklist

- [x] Every contract variable/method declared as public/external private/internal
- [x] Consider if this PR needs the `officeHours` modifier
- [x] Verify expiration (`4 days + 2 hours` monthly and `30 days` for the rest)
- [x] Verify hash in the description matches [here](https://emn178.github.io/online-tools/keccak_256.html)
- [x] Validate all addresses used are in Goerli changelog or known
- [x] Notify any external teams affected by the spell so they have the opportunity to review
- [x] Deploy spell to Goerli `ETH_GAS="XXX" ETH_GAS_PRICE="YYY" make deploy`
- [x] Ensure contract is verified on `Goerli` etherscan
- [x] Change test to use Goerli spell address and deploy timestamp
- [ ] Run `make name="YYYY-MM-DD-archive-directory-name" archive-spell` to make an archive directory and copy `Goerli-DssSpell.sol`, `Goerli-DssSpell.t.sol`, and `Goerli-DssSpell.t.base.sol` 


src/Goerli-DssSpell.sol
```diff
@@ -20,72 +20,70 @@ pragma experimental ABIEncoderV2;
...
-            kprFlatReward:         300,         // 300 Dai
-            kprPctReward:          10           // 0.1%
-        }));
-        DssExecLib.setStairstepExponentialDecrease(MCD_CLIP_CALC_WSTETH_A, 90 seconds, 9900);
-        DssExecLib.setIlkAutoLineParameters("WSTETH-A", 50 * MILLION, 3 * MILLION, 8 hours);
+    address constant WBTC                   = 0x7ccF0411c7932B99FC3704d68575250F032e3bB7;
+    address constant MCD_JOIN_WBTC_B        = 0x13B8EB3d2d40A00d65fD30abF247eb470dDF6C25;
+    address constant MCD_CLIP_WBTC_B        = 0x4F51B15f8B86822d2Eca8a74BB4bA1e3c64F733F;
+    address constant MCD_CLIP_CALC_WBTC_B   = 0x1b5a9aDaf15CAE0e3d0349be18b77180C1a0deCc;
+    address constant PIP_WBTC               = 0xE7de200a3a29E9049E378b52BD36701A0Ce68C3b;
```
gbalabasquer, line 48: Get this from the changelog

src/Goerli-DssSpell.sol
```diff
@@ -20,72 +20,70 @@ pragma experimental ABIEncoderV2;
...
-            breakerTolerance:      5000,        // Allows for a 50% hourly price drop before disabling liquidations
-            auctionDuration:       140 minutes,
-            permittedDrop:         4000,        // 40% price drop before reset
-            liquidationRatio:      16000,       // 160% collateralization
-            kprFlatReward:         300,         // 300 Dai
-            kprPctReward:          10           // 0.1%
-        }));
-        DssExecLib.setStairstepExponentialDecrease(MCD_CLIP_CALC_WSTETH_A, 90 seconds, 9900);
-        DssExecLib.setIlkAutoLineParameters("WSTETH-A", 50 * MILLION, 3 * MILLION, 8 hours);
+    address constant WBTC                   = 0x7ccF0411c7932B99FC3704d68575250F032e3bB7;
```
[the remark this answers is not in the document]
naszam: Done.

src/Goerli-DssSpell.sol
naszam: Done.

src/Goerli-DssSpell.sol
```diff
@@ -20,72 +20,79 @@ pragma experimental ABIEncoderV2;
...
-            liquidationRatio:      16000,       // 160% collateralization
-            kprFlatReward:         300,         // 300 Dai
-            kprPctReward:          10           // 0.1%
-        }));
-        DssExecLib.setStairstepExponentialDecrease(MCD_CLIP_CALC_WSTETH_A, 90 seconds, 9900);
-        DssExecLib.setIlkAutoLineParameters("WSTETH-A", 50 * MILLION, 3 * MILLION, 8 hours);
+    constructor() public {
+        WBTC = DssExecLib.getChangelogAddress("WBTC");
+        PIP_WBTC = DssExecLib.getChangelogAddress("PIP_WBTC");
+    }
```
gbalabasquer, line 56: The immutable thing is actually a good hack in case we get stack too deep. We will definitely need it if NEXO thing is happening. Otherwise we could just get them as regular variables inside the `actions()` (the regular thing we do).
Anyway let's leave them as they are for now, if we confirm that the spell won't be that big we can do the change.

src/Goerli-DssSpell.sol
```diff
@@ -20,72 +20,79 @@ pragma experimental ABIEncoderV2;
...
+            CollateralOpts({
+                ilk:                   "WBTC-B",
+                gem:                   WBTC,
+                join:                  MCD_JOIN_WBTC_B,
+                clip:                  MCD_CLIP_WBTC_B,
+                calc:                  MCD_CLIP_CALC_WBTC_B,
+                pip:                   PIP_WBTC,
+                isLiquidatable:        true,
+                isOSM:                 true,
+                whitelistOSM:          false,
```
gbalabasquer, line 75: It is fine to leave this one in false, as this is already set. But honestly it might be confusing for the future as a new collateral with a first used token with a median + osm needs this on `true`.
The flag itself is thought to put in `false` when the `osm` source is a `ds-value` instead of a median, which actually never happened.

gbalabasquer: I think we might want to revisit a bit these flags but for now I'd just put in `true`.
cc @brianmcmichael

src/Goerli-DssSpell.sol
```diff
@@ -20,72 +20,79 @@ pragma experimental ABIEncoderV2;
...
+                ilkDebtCeiling:        500 * MILLION,
```
gbalabasquer, line 76: This should be set to 30M.

src/Goerli-DssSpell.sol
```diff
@@ -20,72 +20,79 @@ pragma experimental ABIEncoderV2;
...
+                minVaultAmount:        30000,
+                maxLiquidationAmount:  25 * MILLION,
+                liquidationPenalty:    1300,           // 13% penalty fee
+                ilkStabilityFee:       SEVEN_PCT_RATE, // 7% stability fee
+                startingPriceFactor:   12000,          // Auction price begins at 130% of oracle
```
gbalabasquer, line 81:
```suggestion
                startingPriceFactor:   12000,          // Auction price begins at 120% of oracle
```

src/Goerli-DssSpell.sol
gbalabasquer:
```suggestion
                ilkDebtCeiling:        30 * MILLION,
```

src/Goerli-DssSpell.t.base.sol
```diff
@@ -2071,7 +2100,8 @@ contract GoerliDssSpellTestBase is DSTest, DSMath {
 
         (,,,, uint256 dust) = vat.ilks(_ilk);
         dust /= RAY;
-        uint256 amount = 2 * dust * WAD / (_isOSM ? getOSMPrice(pip) : uint256(DSValueAbstract(pip).read()));
+        uint256 amount = 2 * dust * 10**token.decimals() / (_isOSM ? getOSMPrice(pip) : uint256(DSValueAbstract(pip).read()));
+        uint256 amount18 = token.decimals() == 18 ? amount : amount * 10**(18 - token.decimals());
```
gbalabasquer, line 2104: I think is better you leave the math as it was and rename that variable to `amount18` and then divide that value by 10 ^ (decimals difference). It should be better for rounding when dividing by the price.

src/Goerli-DssSpell.sol
```diff
@@ -50,11 +52,28 @@ contract DssSpellAction is DssAction {
...
+    uint256 constant CURRENT_COMP_MAT       =  165 * RAY / 100;
+
+    // --- Offboarding: Target Liquidation Ratio ---
+    uint256 constant TARGET_AAVE_MAT        = 1000 * RAY / 100;
+    uint256 constant TARGET_BAL_MAT         = 1600 * RAY / 100;
+    uint256 constant TARGET_COMP_MAT        =  500 * RAY / 100;
+
     constructor() public {
         WBTC = DssExecLib.getChangelogAddress("WBTC");
         PIP_WBTC = DssExecLib.getChangelogAddress("PIP_WBTC");
```
gbalabasquer: Let's move this two inside the `actions()` as a regular variable.

src/Goerli-DssSpell.t.base.sol
```diff
@@ -947,17 +947,17 @@ contract GoerliDssSpellTestBase is DSTest, DSMath {
...
         });
         afterSpell.collaterals["AAVE-A"] = CollateralValues({
-            aL_enabled:   true,
-            aL_line:      50 * MILLION,
-            aL_gap:       5 * MILLION,
-            aL_ttl:       8 hours,
+            aL_enabled:   false,
+            aL_line:      0,
+            aL_gap:       0,
+            aL_ttl:       0 hours,
```
gbalabasquer, line 953: I'd remove the `hours` like in the other two.

src/Goerli-DssSpell.t.base.sol
```diff
@@ -2294,6 +2294,28 @@ contract GoerliDssSpellTestBase is DSTest, DSMath {
...
+        scheduleWaitAndCast(address(spell));
+        assertTrue(spell.done());
+
+        LerpAbstract lerp = LerpAbstract(lerpFactory.lerps(_lerp));
+
+        hevm.warp(block.timestamp + lerp.duration() / 2);
+        assertEq(getMat(_ilk), _startMat * RAY / 100);
+        lerp.tick();
+        assertGt(getMat(_ilk), _lowMidMat * RAY / 100);
+        assertLt(getMat(_ilk), _highMidMat * RAY / 100);
```
gbalabasquer, line 2312: There is an `assertEqAprox` in the charter PR you could copy and paste.
Regardless of which function you use I'd suggest making the range tighter.

src/Goerli-DssSpell.t.sol
```diff
@@ -83,6 +84,16 @@ contract DssSpellTest is GoerliDssSpellTestBase {
...
 
+    function testAAVELerpOffboardings() public {
+        checkIlkLerpOffboarding("AAVE-A", "AAVE Offboarding", 165, 500, 600, 1000);
+    }
+    function testBALLerpOffboardings() public {
+        checkIlkLerpOffboarding("BAL-A", "BAL Offboarding", 165, 800, 1000, 1600);
+    }
+    function testCOMPLerpOffboardings() public {
+        checkIlkLerpOffboarding("COMP-A", "COMP Offboarding", 165, 250, 350, 500);
+    }
```
gbalabasquer: I'd suggest making just one test for these ones. Taking into account they are the same subject makes sense to group them and will make the run faster.

src/Goerli-DssSpell.t.base.sol
```diff
@@ -1875,7 +1875,7 @@ contract GoerliDssSpellTestBase is DSTest, DSMath {
                 // Convert whole Dai units to expected RAD
                 uint256 normalizedTesthole = values.collaterals[ilk].dog_hole * RAD;
                 assertEq(hole, normalizedTesthole, string(abi.encodePacked("TestError/dog-hole-", ilk)));
-                assertTrue(hole == 0 || hole >= RAD && hole <= 50 * MILLION * RAD, string(abi.encodePacked("TestError/dog-hole-range-", ilk)));
+                assertTrue(hole == 0 || hole >= RAD && hole <= 65 * MILLION * RAD, string(abi.encodePacked("TestError/dog-hole-range-", ilk)));
```
gbalabasquer, line 1878: I think you can make it like 100M so we are fine for a while.

src/Goerli-DssSpell.t.base.sol
gbalabasquer: I'll check this one in detail according to what you share in chat, for now just ignore it.

src/Goerli-DssSpell.sol
naszam: Done in e023fa6.

src/Goerli-DssSpell.t.base.sol
naszam: Done in 832c7e7.

src/Goerli-DssSpell.t.base.sol
naszam: Done in b82e02d.

src/Goerli-DssSpell.t.base.sol
naszam: Done in 7bbbdc0.

src/Goerli-DssSpell.sol
naszam: Done in e0d4f5f.

src/Goerli-DssSpell.t.base.sol
```diff
@@ -2308,8 +2325,7 @@ contract GoerliDssSpellTestBase is DSTest, DSMath {
         hevm.warp(block.timestamp + lerp.duration() / 2);
         assertEq(getMat(_ilk), _startMat * RAY / 100);
         lerp.tick();
-        assertGt(getMat(_ilk), _lowMidMat * RAY / 100);
-        assertLt(getMat(_ilk), _highMidMat * RAY / 100);
+        assertEqApprox(getMat(_ilk), _midMat * RAY / 100, _gapMidMat * RAY / 100);
```
gbalabasquer, line 2326: Why do you really need this midMat and gapMidMat?
I think you can just make the (start + end) / 2 and in tolerance just put a few weis do rounding.

src/Goerli-DssSpell.t.sol
```diff
@@ -86,13 +86,13 @@ contract DssSpellTest is GoerliDssSpellTestBase {
     }
 
     function testAAVELerpOffboardings() public {
-        checkIlkLerpOffboarding("AAVE-A", "AAVE Offboarding", 165, 1100, 1200, 2200);
+        checkIlkLerpOffboarding("AAVE-A", "AAVE Offboarding", 165, 1100, 83, 2200); // 83% tollerance
```
gbalabasquer, line 89: And 83% tolerance? I think I'm missing something about this.

src/Goerli-DssSpell.t.base.sol
gbalabasquer: Ok yeah it's fine as it is because it will truncate the value by the token decimals precision when doing the join.

gbalabasquer (approved):
As soon as the polls are confirmed this looks good to me to deploy and execute.

src/Goerli-DssSpell.t.base.sol
```diff
@@ -2328,6 +2337,31 @@ contract GoerliDssSpellTestBase is DSTest, DSMath {
...
+        vote(address(spell));
+        scheduleWaitAndCast(address(spell));
+        assertTrue(spell.done());
+
+        LerpFactoryAbstract OLD_LERP_FAB = LerpFactoryAbstract(0xbBD821c291c492c40Db2577D9b6E5B1bdAEBD207);
+        LerpAbstract oldLerp = LerpAbstract(OLD_LERP_FAB.lerps(_oldLerp));
+
+        uint256 t = (block.timestamp - oldLerp.startTime()) * WAD / oldLerp.duration();
+        uint256 tickMat = oldLerp.end() * t / WAD + oldLerp.start() - oldLerp.start() * t / WAD;
+        assertEq(getMat(_ilk), tickMat);
```
gbalabasquer, line 2353: This might be avoided, it is actually testing something that already existed and was already proved it worked. I know it doesn't really harm to have more checks than needed but also makes the test larger.

gbalabasquer: Up to you what to do.

src/Goerli-DssSpell.t.base.sol
```diff
@@ -2328,6 +2337,31 @@ contract GoerliDssSpellTestBase is DSTest, DSMath {
...
+        LerpAbstract newLerp = LerpAbstract(lerpFactory.lerps(_newLerp));
+
+        hevm.warp(block.timestamp + newLerp.duration() / 2);
+        assertEq(getMat(_ilk), tickMat);
+        newLerp.tick();
+        assertEqApprox(getMat(_ilk), (tickMat + _newEndMat * RAY / 100) / 2, RAY / 100);
+
+        hevm.warp(block.timestamp + newLerp.duration());
+        newLerp.tick();
+        assertEq(getMat(_ilk), _newEndMat * RAY / 100);
```
gbalabasquer, line 2365: Instead of this you could actually call the regular test.
So to be honest I would use the regular function + checking the old lerp is denied. That could be handled directly in the test and you avoid adding a new base function.

gbalabasquer (approved):
LGTM

PR #33, opened 2021-12-07
PE-771: 2021-12-08 Görli executive spell

julienmartinlevrai:
# Description

# Contribution Checklist

- [x] PR title starts with `(PE-<TICKET_NUMBER>)`
- [ ] Code approved
- [ ] Tests approved
- [ ] CI Tests pass

# Checklist

- [ ] Every contract variable/method declared as public/external private/internal
- [ ] Consider if this PR needs the `officeHours` modifier
- [ ] Verify expiration (`4 days + 2 hours` monthly and `30 days` for the rest)
- [ ] Verify hash in the description matches [here](https://emn178.github.io/online-tools/keccak_256.html)
- [ ] Validate all addresses used are in Goerli changelog or known
- [ ] Notify any external teams affected by the spell so they have the opportunity to review
- [ ] Deploy spell to Goerli `ETH_GAS="XXX" ETH_GAS_PRICE="YYY" make deploy`
- [ ] Ensure contract is verified on `Goerli` etherscan
- [ ] Change test to use Goerli spell address and deploy timestamp
- [ ] Run `make name="YYYY-MM-DD-archive-directory-name" archive-spell` to make an archive directory and copy `Goerli-DssSpell.sol`, `Goerli-DssSpell.t.sol`, and `Goerli-DssSpell.t.base.sol` 
- [ ] `squash and merge` this PR

PR #34, opened 2021-12-17
Update dss interfaces

iamchrissmith:
Add dss-interfaces and update dss-exec-lib so we can use `WardsAbstract` and `PsmAbstract` from there


naszam (approved):
LGTM

iamchrissmith:
waiting until #33 is merged

brianmcmichael (approved):

PR #41, opened 2022-01-19
`cast-dssspell.sh`

naszam:
# Description

# Contribution Checklist

- [ ] PR title starts with `(PE-<TICKET_NUMBER>)`
- [ ] Code approved
- [ ] Tests approved
- [ ] CI Tests pass

# Checklist

- [ ] Every contract variable/method declared as public/external private/internal
- [ ] Consider if this PR needs the `officeHours` modifier
- [ ] Verify expiration (`4 days + 2 hours` monthly and `30 days` for the rest)
- [ ] Verify hash in the description matches [here](https://emn178.github.io/online-tools/keccak_256.html)
- [ ] Validate all addresses used are in Goerli changelog or known
- [ ] Notify any external teams affected by the spell so they have the opportunity to review
- [ ] Deploy spell to Goerli `ETH_GAS="XXX" ETH_GAS_PRICE="YYY" make deploy`
- [ ] Ensure contract is verified on `Goerli` etherscan
- [ ] Change test to use Goerli spell address and deploy timestamp
- [ ] Run `make name="YYYY-MM-DD-archive-directory-name" archive-spell` to make an archive directory and copy `Goerli-DssSpell.sol`, `Goerli-DssSpell.t.sol`, and `Goerli-DssSpell.t.base.sol` 
- [ ] `squash and merge` this PR

PR #45, opened 2022-01-28
`wards.sh`

naszam:
# Description

# Contribution Checklist

- [ ] Code approved
- [ ] Tests approved
- [ ] CI Tests pass

# Checklist

- [ ] Deploy spell to Goerli `ETH_GAS="XXX" ETH_GAS_PRICE="YYY" make deploy`
- [ ] Run `make archive-spell` or `make date="YYYY-MM-DD" archive-spell` to make an archive directory and copy `Goerli-DssSpell.sol`, `Goerli-DssSpell.t.sol`, `Goerli-DssSpell.t.base.sol`, and `Goerli-DssSpellCollateralOnboarding.sol`

### DIFF
--- a/wards.sh
+++ b/wards.sh
@@ -5,6 +5,10 @@ set -e
 
 [[ "$1" ]] || { echo "Please specify the ChainLog Key in ASCII to inspect"; exit 1; }
 
+### Override maxFeePerGas to avoid spikes
+[[ $ETH_GAS_PRICE && $ETH_GAS_PRICE -lt $(seth basefee) ]]
+export ETH_GAS_PRICE=$((ETH_GAS_PRICE + $(seth --to-wei 200 gwei)))
+
 CHANGELOG=0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F
 KEY=$(seth --to-bytes32 "$(seth --from-ascii "$1")")
 TARGET=$(seth call "$CHANGELOG" 'getAddress(bytes32)(address)' "$KEY")
--- a/wards.sh
+++ b/wards.sh
@@ -8,7 +8,7 @@ set -e
 ### Override maxFeePerGas to avoid spikes
 BASE_FEE=$(seth basefee)
 if [ -n "$ETH_GAS_PRICE" ] && [ "$ETH_GAS_PRICE" -lt "$BASE_FEE" ]; then
-        ETH_GAS_PRICE=$((BASE_FEE * 3))
+        export ETH_GAS_PRICE=$((BASE_FEE * 3))
 fi
 
 CHANGELOG=0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F
--- a/wards.sh
+++ b/wards.sh
@@ -7,10 +7,8 @@ set -e
 
 ### Override maxFeePerGas to avoid spikes
 BASE_FEE=$(seth basefee)
-ethGasPriceLtBaseFee=$(echo "$ETH_GAS_PRICE < $BASE_FEE" | bc)
-if [[ -n "$ETH_GAS_PRICE" && "$ethGasPriceLtBaseFee" == 1 ]]; then
-        export ETH_GAS_PRICE=$(echo "$BASE_FEE * 3" | bc)
-fi
+[[ -n "$ETH_GAS_PRICE" ]] && ethGasPriceLtBaseFee=$(echo "$ETH_GAS_PRICE < $BASE_FEE" | bc)
+[[ "$ethGasPriceLtBaseFee" == 1 ]] && export "ETH_GAS_PRICE=$(echo "$BASE_FEE * 3" | bc)"
 
 CHANGELOG=0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F
 
--- a/wards.sh
+++ b/wards.sh
@@ -3,7 +3,7 @@ set -e
 
 [[ "$ETH_RPC_URL" ]] || { echo "Please set a ETH_RPC_URL"; exit 1; }
 
-[[ "$1" ]] || { echo "Please specify the ChainLog Key in ASCII to inspect"; exit 1; }
+[[ "$1" ]] || { echo "Please specify the Target Address or ChainLog Key (ASCII) to inspect"; exit 1; }
 
 ### Override maxFeePerGas to avoid spikes
 BASE_FEE=$(seth basefee)
--- a/wards.sh
+++ b/wards.sh
@@ -13,21 +13,21 @@ BASE_FEE=$(seth basefee)
 CHANGELOG=0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F
 
 if [[ "$1" =~ 0x* ]]; then
-        TARGET=$1
+    TARGET=$1
 else
-        KEY=$(seth --to-bytes32 "$(seth --from-ascii "$1")")
-        TARGET=$(seth call "$CHANGELOG" 'getAddress(bytes32)(address)' "$KEY")
+    KEY=$(seth --to-bytes32 "$(seth --from-ascii "$1")")
+    TARGET=$(seth call "$CHANGELOG" 'getAddress(bytes32)(address)' "$KEY")
 fi
 
 echo -e "Network: $(seth chain)"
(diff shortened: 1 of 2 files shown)


wards.sh
```diff
@@ -0,0 +1,19 @@
...
+
+CHANGELOG=0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F
+KEY=$(seth --to-bytes32 "$(seth --from-ascii "$1")")
+TARGET=$(seth call "$CHANGELOG" 'getAddress(bytes32)(address)' "$KEY")
+
+echo -e "Network: $(seth chain)"
+LIST=$(seth call "$CHANGELOG" 'list()(bytes32[])')
+for key in $(echo -e "$LIST" | sed "s/,/ /g")
+do
+        ADDRESS=$(seth call "$CHANGELOG" 'getAddress(bytes32)(address)' "$key")
```
brianmcmichael, line 16: Getting gas errors here because mainnet fee spiked. Might want to set the gas envvar to some high number in the script so it doesn't affect reads.

```haskell
bmcmichael@bmcx1e:~/makerdao/spells-goerli$ ./wards.sh "MCD_ESM"
seth-rpc: {"id":1,"jsonrpc":"2.0","method":"eth_call","params":[{"data":"0x21f8a7214d43445f45534d00000000000000000000000000000000000000000000000000","from":"0xC0dE1AB51082442f821bA273273B4F7d32521D8b","gas":"0x895440","maxFeePerGas":"0x2e90edd000","maxPriorityFeePerGas":"0x77359400","to":"0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F"},"latest"]}
seth-rpc: error:   code       -32000
seth-rpc: error:   message    err: max fee per gas less than block base fee: address 0xabc..., maxFeePerGas: 200000000000 baseFee: 206037515885 (supplied gas 9000000)
```

naszam: Haven't reproduced it but d0709f8 should fix it.
Let me know if that works or some more tweaks are needed.

wards.sh
```diff
@@ -5,6 +5,10 @@ set -e
 
 [[ "$1" ]] || { echo "Please specify the ChainLog Key in ASCII to inspect"; exit 1; }
 
+### Override maxFeePerGas to avoid spikes
+[[ $ETH_GAS_PRICE && $ETH_GAS_PRICE -lt $(seth basefee) ]]
+export ETH_GAS_PRICE=$((ETH_GAS_PRICE + $(seth --to-wei 200 gwei)))
```
brianmcmichael, line 10: I would say don't `export` this, just set it like a normal variable so it's scoped to and doesn't escape the script. 

You might also consider setting the ETH_GAS_PRICE just double or triple the `seth basefee` here so that it works in high gas environments.

naszam: done in ccbd004

brianmcmichael (approved):
LGTM!

wards.sh
```diff
@@ -0,0 +1,25 @@
...
+CHANGELOG=0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F
+KEY=$(seth --to-bytes32 "$(seth --from-ascii "$1")")
+TARGET=$(seth call "$CHANGELOG" 'getAddress(bytes32)(address)' "$KEY")
+
+echo -e "Network: $(seth chain)"
+LIST=$(seth call "$CHANGELOG" 'list()(bytes32[])')
+for key in $(echo -e "$LIST" | sed "s/,/ /g")
+do
+        ADDRESS=$(seth call "$CHANGELOG" 'getAddress(bytes32)(address)' "$key")
+        WARDS=$(seth call "$ADDRESS" 'wards(address)(uint256)' "$TARGET" 2>/dev/null) || continue
```
julienmartinlevrai, line 23: I would suggest adding an echo here to show the progress being made.

julienmartinlevrai:
We are assuming that the chainlog stores all the relevant addresses we need to check wards against. That is a heavy assumption because there are some contracts which, unfortunately, are not in the chainlog and are still relevant. In particular, medianizers come to mind.

A way to include medianizers in the check would be to call `src()` on every chainlog contract and if it returns an address, add it also to the check.

wards.sh
naszam: this line tries to simulate the low level call pattern used by `checkWards` in spell base tests.
the `	WARDS` var and the following conditional block have been added to only show the chainlog entries that relies the target contract to reduce verbosity.

julienmartinlevrai: Good point. It is a bit confusing at first, but it makes sense to only show the addresses. Resolving

wards.sh
gbalabasquer: Actually you need to export, otherwise it won't be used in the consequent `seth` calls.
It won't escape the script.
The alternative is you pass it on every `seth` call like: `ETH_GAS=$ETH_GAS seth send...`

gbalabasquer: Just to clarify the `export` won't do it one upper level of the script but it will make it a global variable for the whole script.

wards.sh
naszam: I've adopted `bc` to handle big number in 81201f8

naszam:
> We are assuming that the chainlog stores all the relevant addresses we need to check wards against. That is a heavy assumption because there are some contracts which, unfortunately, are not in the chainlog and are still relevant. In particular, medianizers come to mind.
> 
> A way to include medianizers in the check would be to call `src()` on every chainlog contract and if it returns an address, add it also to the check.

The `wards` script was crafted to inspect current chainlog against the old ESM, inspired by the integration test (https://github.com/makerdao/spells-goerli/blob/master/archive/2022-01-27-DssSpell/Goerli-DssSpell.t.sol#L81) that adopted the `checkWards` pattern used in the base tests.
I agree with you that's not exhaustive, but its main purpose was to inspect ESM auths against current chainlog, to review spell actions.
Following your suggestion, I've extended the script to include the `src` (medianizers) into the list, following the pattern used in `checkSource` in base tests, allowing also to check against a target address outside of the chainlog list (e.g. deployers).
Further extentions can be explored, as suggested by @gbalabasquer, to check against different changelog versions fetching from the UI, following a similar approach as per [build-env-addresses.sh](https://github.com/makerdao/MIP21-RWA-Example/blob/master/scripts/build-env-addresses.sh), in a separate PR eventually.

wards.sh
```diff
@@ -0,0 +1,33 @@
...
+list=$(seth call "$CHANGELOG" 'list()(bytes32[])')
+for key in $(echo -e "$list" | sed "s/,/ /g")
+do
+    contractName=$(seth --to-ascii "$key" | sed 's/\x0/ /g')
+    contract=$(seth call "$CHANGELOG" 'getAddress(bytes32)(address)' "$key")
+    wards=$(seth call "$contract" 'wards(address)(uint256)' "$target" 2>/dev/null) || continue
+    [[ "$wards" == "1" ]] && echo "$contractName"
+    src=$(seth call "$contract" 'src()(address)' 2>/dev/null) || continue
+    srcWards=$(seth call "$src" 'wards(address)(uint256)' "$target" 2>/dev/null) || continue
+    [[ "$srcWards" == "1" ]] && echo "source of $contractName"
```
julienmartinlevrai, line 32:
```suggestion
    [[ "$srcWards" == "1" ]] && echo "source of $contractName ($src)"
```

julienmartinlevrai:
Besides a code suggestion, code looks good

wards.sh
naszam: yeah, I've followed the convention in `checkSource` https://github.com/makerdao/spells-goerli/blob/master/src/Goerli-DssSpell.t.base.sol#L1088, but open to consider a different approach, @brianmcmichael ?

naszam: I would lean to keep it as it is as more descriptive for the end user or apply your suggestions along with a goerli etherscan URL, as done for the spell address in #41, for manual inspection of the source name and wards.

julienmartinlevrai: how about this then
```suggestion
    [[ "$srcWards" == "1" ]] && echo "source of $contractName: $src"
```

naszam: I've included the `src` address in 59d6744, tweaking a bit to address spacing.

julienmartinlevrai (approved):
LGTM

brianmcmichael (approved):

PR #61, opened 2022-03-08
Use `sethSend` in `cast-dssspell.sh`

naszam:
- Use `sethSend` pattern from `dss-deploy-scripts` to address nonce issue in `cast-dssspell.sh`


brianmcmichael (approved):

gbalabasquer (approved):
LGTM

cast-dssspell.sh
```diff
@@ -15,6 +15,16 @@ hat=$(seth call "$MCD_ADM" 'hat()(address)')
 approvals=$(seth call "$MCD_ADM" 'approvals(address)(uint256)' "$hat")
 deposits=$(seth call "$MCD_ADM" 'deposits(address)(uint256)' "$ETH_FROM")
 
+ETH_NONCE=$(seth nonce "$ETH_FROM")
+
+sethSend() {
+    set -e
+    echo "seth send $*"
+    ETH_NONCE="$ETH_NONCE" ETH_GAS=$(seth estimate "$@") seth send "$@"
```
brianmcmichael: I think sometimes this might fail in some cases if we use the exact estimate. Maybe not in spell deployments but sometimes if the on-chain state changes it can affect the cost of a transaction between when we calculate it and when it gets put into a block. We could probably keep this, but we should add maybe 50k to 100k gas in addition to the estimate just to avoid storage slot surprises.

brianmcmichael: Actually, this is probably fine here since this line is just calculating the nonce. Don't worry about the addition.

brianmcmichael (approved):

PR #68, opened 2022-04-20
`time.sh`

naszam:
- Add `time.sh` script to generate and validate timestamps 
	- `make time date="2022-04-20 00:00:00 UTC"`
	- `make time stamp=1650412800`
- Return current date in UTC by default
	- `make time`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,3 +14,4 @@ cast-spell         :; ./scripts/cast-dssspell.sh $(spell)
 archive-spell      :; ./scripts/archive-dssspell.sh "$(if $(date),$(date),$(shell date +'%Y-%m-%d'))"
 diff-archive-spell :; ./scripts/diff-archive-dssspell.sh "$(if $(date),$(date),$(shell date +'%Y-%m-%d'))"
 wards              :; ./scripts/wards.sh $(target)
+time               :; ./scripts/time.sh date="$(date)" stamp="$(stamp)"
--- a/scripts/time.sh
+++ b/scripts/time.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+set -e
+
+for ARGUMENT in "$@"
+do
+    KEY=$(echo "$ARGUMENT" | cut -f1 -d=)
+    VALUE=$(echo "$ARGUMENT" | cut -f2 -d=)
+
+    case "$KEY" in
+            date)      DATE="$VALUE" ;;
+            stamp)     STAMP="$VALUE" ;;
+            *)
+    esac
+done
+
+if [[ -z "$STAMP" && -z "$DATE" ]]; then
+    date --utc # return current date in UTC by default
+elif [[ -z "$DATE" ]]; then
+    date --utc --date=@"$STAMP"
+else
+    date --date "$DATE" +%s
+fi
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -14,3 +14,4 @@ cast-spell         :; ./scripts/cast-dssspell.sh $(spell)
 archive-spell      :; ./scripts/archive-dssspell.sh "$(if $(date),$(date),$(shell date +'%Y-%m-%d'))"
 diff-archive-spell :; ./scripts/diff-archive-dssspell.sh "$(if $(date),$(date),$(shell date +'%Y-%m-%d'))"
 wards              :; ./scripts/wards.sh $(target)
+time               :; ./scripts/time.sh date="$(date)" stamp="$(stamp)"
```
brianmcmichael, line 17: Generally I like this script. I wonder if the current timestamp is useful in our context. It would be nice if `date` or `stamp` is not set the script echoes some help to using the proper format? i.e.:

```
$ make time
> Please use the date or stamp flag, ex:
>    make time date="5/1/22 UTC" 
>       or
>    make time stamp="1651363200"
```

I, for one, know that I will probably forget how to format the date string.

brianmcmichael: Also I noticed that "date" by itself doesn't convert to UTC by default so users will need to be specific. Example:

```
bmcmichael@bmcx1e:~/makerdao/spells-goerli$ make time date="5/1/22"
./scripts/time.sh date="5/1/22" stamp=""
1651377600
bmcmichael@bmcx1e:~/makerdao/spells-goerli$ make time stamp="1651377600"
./scripts/time.sh date="" stamp="1651377600"
Sun 01 May 2022 04:00:00 AM UTC
```

julienmartinlevrai:
@naszam can you please elaborate on the use case for this script ? I might have missed it.

naszam:
> @naszam can you please elaborate on the use case for this script ? I might have missed it.

DssVest streams, to generate and inspect timestamps.

Makefile
naszam: I've added usage message instead of current date in 65ab0ba.

julienmartinlevrai: Yes, it would be better if users didn’t need to specify UTC when they enter a date. I think currently the scripts assumes the user’s local time.

julienmartinlevrai: Since users can input dates in many different formats when using the `date` flag, it would be better to not only output the timestamp, but also the date string just so that the user makes sure it is the intended date.

naszam: I've applied the `--utc` option to omit the `UTC` frome the date string in 44647d6

brianmcmichael: Still having an issue with this. It looks like the "date" flag still isn't respecting UTC. I made some local mods to the output and it's still off:

```
bmcmichael@bmcx1e:~/makerdao/spells-goerli$ make time date="6/12/22"
./scripts/time.sh date="6/12/22" stamp=""
Sun 12 Jun 2022 12:00:00 AM EDT: 1655006400
bmcmichael@bmcx1e:~/makerdao/spells-goerli$ make time stamp="1655006400"
./scripts/time.sh date="" stamp="1655006400"
Sun 12 Jun 2022 04:00:00 AM UTC: 1655006400
```

julienmartinlevrai (approved):
LGTM

brianmcmichael (approved):

PR #69, opened 2022-04-25
Fix bug in test when new chainlog entries are added and version is changed

brianmcmichael:
Fixes the case where new entries are added to the changelog and the version is appropriately updated by short-circuiting the test.

### DIFF
--- a/src/Goerli-DssSpell.t.sol
+++ b/src/Goerli-DssSpell.t.sol
@@ -447,6 +447,9 @@ contract DssSpellTest is GoerliDssSpellTestBase {
             if (keccak256(abi.encodePacked(_version)) == keccak256(abi.encodePacked(chainLog.version()))) {
                 emit log_named_string("Error", concat("TestError/chainlog-version-not-updated-count-change-", _version));
                 fail();
+            } else {
+                // Chainlog size changed and version is updated, so test passes.
+                return;
             }
         }
 
--- a/src/Goerli-DssSpell.t.sol
+++ b/src/Goerli-DssSpell.t.sol
@@ -447,7 +447,7 @@ contract DssSpellTest is GoerliDssSpellTestBase {
             // Chainlog version has been updated. Short circuit the test here.
             return;
         } else {
-            // Fail if the version is not updated and the chainlog count has not changed
+            // Fail if the version is not updated and the chainlog count has changed
             if (_count != chainLog.count()) {
                 emit log_named_string("Error", concat("TestError/chainlog-version-not-updated-count-change-", _version));
                 fail();
--- a/src/Goerli-DssSpell.t.sol
+++ b/src/Goerli-DssSpell.t.sol
@@ -443,10 +443,7 @@ contract DssSpellTest is GoerliDssSpellTestBase {
         scheduleWaitAndCast(address(spell));
         assertTrue(spell.done());
 
-        if (keccak256(abi.encodePacked(_version)) != keccak256(abi.encodePacked(chainLog.version()))) {
-            // Chainlog version has been updated. Short circuit the test here.
-            return;
-        } else {
+        if (keccak256(abi.encodePacked(_version)) == keccak256(abi.encodePacked(chainLog.version()))) {
             // Fail if the version is not updated and the chainlog count has changed
             if (_count != chainLog.count()) {
                 emit log_named_string("Error", concat("TestError/chainlog-version-not-updated-count-change-", _version));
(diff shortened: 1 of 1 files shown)


src/Goerli-DssSpell.t.sol
```diff
@@ -443,23 +443,23 @@ contract DssSpellTest is GoerliDssSpellTestBase {
         scheduleWaitAndCast(address(spell));
         assertTrue(spell.done());
 
-        if (_count != chainLog.count()) {
-            if (keccak256(abi.encodePacked(_version)) == keccak256(abi.encodePacked(chainLog.version()))) {
+        if (keccak256(abi.encodePacked(_version)) != keccak256(abi.encodePacked(chainLog.version()))) {
+            // Chainlog version has been updated. Short circuit the test here.
+            return;
+        } else {
```
gbalabasquer, line 449: Maybe:
```suggestion
        if (keccak256(abi.encodePacked(_version)) == keccak256(abi.encodePacked(chainLog.version()))) {
```
?

src/Goerli-DssSpell.t.sol
```diff
@@ -443,23 +443,23 @@ contract DssSpellTest is GoerliDssSpellTestBase {
...
-
-        for(uint256 i = 0; i < chainLog.count(); i++) {
-            (, address _val) = chainLog.get(i);
-            // If the address arrays don't match it's due to a change in the changelog. Fail if the version is not updated.
-            if (_chainlog_addrs[i] != _val) {
-                if (keccak256(abi.encodePacked(_version)) == keccak256(abi.encodePacked(chainLog.version()))) {
+            // Fail if the chainlog is the same size but local keys don't match the chainlog.
+            for(uint256 i = 0; i < chainLog.count(); i++) {
+                (, address _val) = chainLog.get(i);
+                if (address(_chainlog_addrs[i]) != address(_val)) {
```
gbalabasquer, line 458:
```suggestion
                if (_chainlog_addrs[i] != _val) {
```

src/Goerli-DssSpell.t.sol
brianmcmichael: Much better.

brianmcmichael: fixed

src/Goerli-DssSpell.t.sol
```diff
@@ -443,23 +443,20 @@ contract DssSpellTest is GoerliDssSpellTestBase {
...
             }
-        }
-
-        for(uint256 i = 0; i < chainLog.count(); i++) {
-            (, address _val) = chainLog.get(i);
-            // If the address arrays don't match it's due to a change in the changelog. Fail if the version is not updated.
-            if (_chainlog_addrs[i] != _val) {
-                if (keccak256(abi.encodePacked(_version)) == keccak256(abi.encodePacked(chainLog.version()))) {
+            // Fail if the chainlog is the same size but local keys don't match the chainlog.
+            for(uint256 i = 0; i < chainLog.count(); i++) {
```
gbalabasquer, line 453:
```suggestion
            for(uint256 i = 0; i < _count; i++) {
```
Just to be annoying lol

src/Goerli-DssSpell.t.sol
```diff
@@ -443,23 +443,20 @@ contract DssSpellTest is GoerliDssSpellTestBase {
...
+                (, address _val) = chainLog.get(i);
```
gbalabasquer, line 454: I think it would be better to bring the whole array and then compare one registry of each memory array. That will reduce drastically the calls to the node.

src/Goerli-DssSpell.t.sol
gbalabasquer: Actually I think it would be better to call the whole array as the original one and compare each memory array element against the other. That will reduce drastically the calls to the node.

brianmcmichael: That makes sense. I'll make the change.

brianmcmichael: I don't think we'll see a reduction in calls here since there's a separate loop to populate the local array:

```
for(uint256 i = 0; i < _count; i++) {
            (, address _val) = chainLog.get(i);
            _chainlog_addrs[i] = _val;
        }
```

This would need to be replicated on the other side of the cast.

gbalabasquer: Nvm, I was just mixing stuff with the `list` method which only brings the keys. Then the only one that can be saved in the extra count of the first comment.

gbalabasquer (approved):
LGTM when adding the return after fail()

naszam (approved):
LGTM

PR #87, opened 2022-06-23
add starknet tests

brianmcmichael:
# Description

Adds starknet tests to goerli.

Not sure I love this design. Wanted to keep the starknet stuff separate from everything else and encapsulated into their own file because it makes sense to keep some of this external stuff out of base tests, but open to options on how to streamline this as well.


brianmcmichael:
> Is it to avoid repeating the spell cast code ?

Yes. Since all of these checks take place after the spell cast we only need to simulate one schedule/vote/cast.

This might not be as relevant an optimization after the transition to forge, but the idea was that it would limit the number of cycles/rpc calls in processing these tests.

PR #85, opened 2022-06-21
PE-1026-20220624: Weekly Executive (Görli)

julienmartinlevrai:
# Description

# Contribution Checklist

- [x] PR title starts with `(PE-<TICKET_NUMBER>)`
- [ ] Code approved
- [ ] Tests approved
- [ ] CI Tests pass

# Checklist

- [ ] Every contract variable/method declared as public/external private/internal
- [ ] Consider if this PR needs the `officeHours` modifier override
- [ ] Verify expiration (`30 days` unless otherwise specified)
- [ ] Verify hash in the description matches [here](https://emn178.github.io/online-tools/keccak_256.html)
- [ ] Validate all addresses used are in Goerli changelog or known
- [ ] Notify any external teams affected by the spell so they have the opportunity to review
- [ ] Deploy spell to Goerli `ETH_GAS="XXX" ETH_GAS_PRICE="YYY" make deploy`
- [ ] Ensure contract is verified on `Goerli` etherscan
- [ ] Change test to use Goerli spell address and deploy timestamp
- [ ] Cast spell on Goerli `make spell="0x-deployed-spell-address" cast-spell`
- [ ] Run `make archive-spell` or `make date="YYYY-MM-DD" archive-spell` to make an archive directory and copy `Goerli-DssSpell.sol`, `Goerli-DssSpell.t.sol`, `Goerli-DssSpell.t.base.sol`, and `Goerli-DssSpellCollateralOnboarding.sol`
- [ ] `squash and merge` this PR

PR #90, opened 2022-07-05
Update spell templates

naszam:
- Rename collateral onboarding contract following #85 
- Update `onboardNewCollaterals` template
- Include `offboardCollaterals` action
- Update IPFS Rates Table Link
- Cleanups


template/Goerli-DssSpellCollateral.sol
```diff
@@ -0,0 +1,104 @@
...
+    // uint256 constant BILLION    = 10 ** 9;
+
+    // --- DEPLOYED COLLATERAL ADDRESSES ---
+    // address constant XXX                  = 0x0000000000000000000000000000000000000000;
+    // address constant PIP_XXX              = 0x0000000000000000000000000000000000000000;
+    // address constant MCD_JOIN_XXX_A       = 0x0000000000000000000000000000000000000000;
+    // address constant MCD_CLIP_XXX_A       = 0x0000000000000000000000000000000000000000;
+    // address constant MCD_CLIP_CALC_XXX_A  = 0x0000000000000000000000000000000000000000;
+
+    function onboardNewCollaterals() internal {
```
julienmartinlevrai, line 48:
```suggestion
    function onboardCollaterals() internal {
```

template/Goerli-DssSpellCollateral.sol
```diff
@@ -0,0 +1,104 @@
...
+        // DssExecLib.setChangelogAddress("XXX", XXX);
+        // DssExecLib.setChangelogAddress("PIP_XXX", PIP_XXX);
+        // DssExecLib.setChangelogAddress("MCD_JOIN_XXX_A", MCD_JOIN_XXX_A);
+        // DssExecLib.setChangelogAddress("MCD_CLIP_XXX_A", MCD_CLIP_XXX_A);
+        // DssExecLib.setChangelogAddress("MCD_CLIP_CALC_XXX_A", MCD_CLIP_CALC_XXX_A);
+    }
+
+    function offboardCollaterals() internal {
+        // ----------------------------- Collateral offboarding -----------------------------
+
```
julienmartinlevrai, line 102: eventually we should add something here

template/Goerli-DssSpell.sol
```diff
@@ -40,17 +42,20 @@ contract DssSpellAction is DssAction, DssSpellCollateralOnboardingAction {
     // $ bc -l <<< 'scale=27; e( l(1.08)/(60 * 60 * 24 * 365) )'
     //
     // A table of rates can be found at
-    //    https://ipfs.io/ipfs/QmefQMseb3AiTapiAKKexdKHig8wroKuZbmLtPLv4u2YwW
+    //    https://ipfs.io/ipfs/QmPgPVrVxDCGyNR5rGp9JC5AUxppLzUAqvncRJDcxQnX1u
```
julienmartinlevrai, line 45: did you use a script to generate this ?

template/Goerli-DssSpellCollateral.sol
julienmartinlevrai: to standardize it with `offboardCollaterals` and to remove the pleonasm

template/Goerli-DssSpell.sol
naszam: Nope, the IPFS hash was taken from `spells-mainnet`

julienmartinlevrai:
Should the archive script copy these to `src` ?

naszam:
> Should the archive script copy these to `src` ?

I think the template could be used during the cleanup phase of the spell (base spell), automating the process via an ad-hoc script eventually

template/Goerli-DssSpell.sol
naszam: But now that I think more on it, it might be nice to add a script to generate rates based on the commented formula to ease the spell crafting and reviewing process and eventually automate IPFS Rates Table updates (that might need some further updates, like including 0.01%).

julienmartinlevrai (approved):
LGTM

template/Goerli-DssSpell.sol
```diff
@@ -40,17 +42,20 @@ contract DssSpellAction is DssAction, DssSpellCollateralOnboardingAction {
     // $ bc -l <<< 'scale=27; e( l(1.08)/(60 * 60 * 24 * 365) )'
     //
     // A table of rates can be found at
-    //    https://ipfs.io/ipfs/QmefQMseb3AiTapiAKKexdKHig8wroKuZbmLtPLv4u2YwW
+    //    https://ipfs.io/ipfs/QmX2QMoM1SZq2XMoTbMak8pZP86Y2icpgPAKDjQg4r4YHn
```
gbalabasquer, line 45: Why is this one changing?

template/Goerli-DssSpellCollateral.sol
```diff
@@ -0,0 +1,168 @@
+// SPDX-FileCopyrightText: © 2021-2022 Dai Foundation <www.daifoundation.org>
```
gbalabasquer, line 1: The best would be to actuall `git mv` the old file to the new name, so that way github still recognizes the history.

naszam: I used `git mv` and committed both `git mv` and updates in the same commit, as mentioned in chat.
I've now undo those changes and repeat the steps separately but It looks the same, where the old file is deleted instead of renamed.

template/Goerli-DssSpell.sol
naszam: I've applied the change in the collater one as well.
It updates with latest IPFS Rates Table (including also 0.01%)

gbalabasquer (approved):

PR #103, opened 2022-08-29
Test cure teleport

talbaneth:


src/Goerli-DssSpell.t.sol
```diff
@@ -565,4 +560,31 @@ contract DssSpellTest is GoerliDssSpellTestBase {
             8 days
         );
     }
+
+    // NOTE: Only executable by forge
```
hexonaut, line 564: This can actually be executed in dapptools just fine.

hexonaut (approved):
Good to merge after comment is removed.

gbalabasquer:
I'm not sure I really like much how these functions with `cage` flag are built and executed. It cost me a bit to understand what was happening. I would think that it is just better to get that extracted out to the test itself and maybe the internal in the base contract wouldn't even be needed at all. But I'm fine if you guys prefer this way. After fixing Sam's comment, should be good to merge.

src/Goerli-DssSpell.t.sol
talbaneth: Done.

PR #115, opened 2022-10-05
port from spells-mainnet

julienmartinlevrai:
The best way to review these changes is to diff them to current spell-mainnet files on master.

### DIFF
--- a/src/test/config.sol
+++ b/src/test/config.sol
@@ -628,7 +628,7 @@ contract Config {
             line:         0,
             dust:         10 * THOUSAND,
             pct:          400,
-            mat:          2430000,
+            mat:          2430000,  // 24,300%
             liqType:      "clip",
             liqOn:        true,
             chop:         0,
--- a/src/test/config.sol
+++ b/src/test/config.sol
@@ -99,7 +99,7 @@ contract Config {
         //
         spellValues = SpellValues({
             deployed_spell:                 address(0xB273Cc14acAF229dAB3eD6AEf83A8bb352919D66),        // populate with deployed spell if deployed
-            deployed_spell_created:         1664905308,                 // use tx=<deploy tx> make deploy-stamp
+            deployed_spell_created:         1664905308,                 // use `./scripts/get-created-timestamp.sh <deployment-tx>`
             previous_spell:                 address(0),        // supply if there is a need to test prior to its cast() function being called on-chain.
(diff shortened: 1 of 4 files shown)


src/test/config.sol
```diff
@@ -99,7 +99,7 @@ contract Config {
         //
         spellValues = SpellValues({
             deployed_spell:                 address(0xB273Cc14acAF229dAB3eD6AEf83A8bb352919D66),        // populate with deployed spell if deployed
```
brianmcmichael, line 101:
```suggestion
            deployed_spell:                 address(0),        // populate with deployed spell if deployed
```

src/test/config.sol
```diff
@@ -99,7 +99,7 @@ contract Config {
...
-            deployed_spell_created:         1664905308,                 // use tx=<deploy tx> make deploy-stamp
+            deployed_spell_created:         1664905308,                 // use `./scripts/get-created-timestamp.sh <deployment-tx>`
```
brianmcmichael, line 102: Does the `make` command not work anymore?

src/test/config.sol
```diff
@@ -628,7 +628,7 @@ contract Config {
             line:         0,
             dust:         10 * THOUSAND,
             pct:          400,
-            mat:          2430000,  // 24,300%
```
brianmcmichael, line 631: Not sure you need to rm this comment

brianmcmichael: It's such a weird number that it the comment helps to indicate that it's intentional

src/test/config.sol
brianmcmichael: nvm the script path is probably better here

src/test/config.sol
julienmartinlevrai: I did this in the spell PR

src/test/config.sol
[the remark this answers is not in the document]
julienmartinlevrai: I did this in the spell PR

src/test/config.sol
julienmartinlevrai: done

iamchrissmith (approved):
looks good to merge

The-Arbiter (commented):
Looks good to me as it is now. Any reason why `testCollateralIntegrations` isn't public by default if you need to make it public to use it? I'm assuming this is for gas savings or that there is some reason for this.

brianmcmichael (approved):
Good to merge

PR #117, opened 2022-10-11
`hash-exec-copy.sh`

naszam:
- Add `hash-exec-copy.sh` script to generate exec hash
	- `make exec-hash url=https://raw.githubusercontent.com/makerdao/community/...`
- Revert to `curl` in case `wget` is not installed
- Sanity check the expected `url` to match

### DIFF
--- a/scripts/hash-exec-copy.sh
+++ b/scripts/hash-exec-copy.sh
@@ -1,12 +1,9 @@
 #!/usr/bin/env bash
 set -e
 
-[[ "$1" =~ https://raw.githubusercontent.com/makerdao/community/* ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/...)"; exit 1; }
+URL=$(echo "$1" | cut -f2 -d=)
 
-for ARGUMENT in "$@"
-do
-    URL=$(echo "$ARGUMENT" | cut -f2 -d=)
-done
+[[ "$URL" == https://raw.githubusercontent.com/makerdao/community/*/governance/votes/*.md ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/<commit>/governance/votes/*.md)"; exit 1; }
 
 if [[ -x "$(command -v wget)" ]]; then
     cast keccak -- "$(wget "$URL" -q -O - 2>/dev/null)"
--- a/scripts/hash-exec-copy.sh
+++ b/scripts/hash-exec-copy.sh
@@ -3,7 +3,7 @@ set -e
 
 URL=$(echo "$1" | cut -f2 -d=)
 
-[[ "$URL" == https://raw.githubusercontent.com/makerdao/community/*/governance/votes/*.md ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/<commit>/governance/votes/*.md)"; exit 1; }
+[[ "$URL" == https://raw.githubusercontent.com/makerdao/community/*/governance/votes/*.md ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/<commit>/governance/votes/<file name>.md)"; exit 1; }
 
 if [[ -x "$(command -v wget)" ]]; then
     cast keccak -- "$(wget "$URL" -q -O - 2>/dev/null)"
--- a/scripts/hash-exec-copy.sh
+++ b/scripts/hash-exec-copy.sh
@@ -1,14 +1,12 @@
 #!/usr/bin/env bash
 set -e
 
-URL=$(echo "$1" | cut -f2 -d=)
-
-[[ "$URL" == https://raw.githubusercontent.com/makerdao/community/*/governance/votes/*.md ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/<commit>/governance/votes/<file name>.md)"; exit 1; }
+[[ "$1" == https://raw.githubusercontent.com/makerdao/community/*/governance/votes/*.md ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/<commit>/governance/votes/<file name>.md)"; exit 1; }
 
 if [[ -x "$(command -v wget)" ]]; then
-    cast keccak -- "$(wget "$URL" -q -O - 2>/dev/null)"
+    cast keccak -- "$(wget "$1" -q -O - 2>/dev/null)"
 elif [[ -x "$(command -v curl)" ]]; then
-    cast keccak -- "$(curl "$URL" -o - 2>/dev/null)"
+    cast keccak -- "$(curl "$1" -o - 2>/dev/null)"
 else
     echo "Please install either wget or curl";
     exit 1;
--- a/scripts/hash-exec-copy.sh
+++ b/scripts/hash-exec-copy.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-[[ "$1" =~ https://raw.githubusercontent.com/makerdao/community/* ]] || { echo "Please provide the correct exec copy link to hash (e.g. https://raw.githubusercontent.com/makerdao/community/...)"; exit 1; }
+[[ "$1" =~ https://raw.githubusercontent.com/makerdao/community/* ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/...)"; exit 1; }
 
 for ARGUMENT in "$@"
 do
--- a/scripts/hash-exec-copy.sh
+++ b/scripts/hash-exec-copy.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-[[ "$1" =~ https://raw.githubusercontent.com/makerdao/community/* ]] || { echo "Plese provide the exec copy link to hash (e.g. url=<https://raw.githubusercontent.com/makerdao/community/...>)"; exit 1; }
+[[ "$1" =~ https://raw.githubusercontent.com/makerdao/community/* ]] || { echo "Please provide the exec copy link to hash (e.g. url=<https://raw.githubusercontent.com/makerdao/community/...>)"; exit 1; }
 
 for ARGUMENT in "$@"
 do
--- a/scripts/hash-exec-copy.sh
+++ b/scripts/hash-exec-copy.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -e
+
+[[ "$1" =~ https://raw.githubusercontent.com/makerdao/community/* ]] || { echo "Plese provide the exec copy link to hash (e.g. url=<https://raw.githubusercontent.com/makerdao/community/...>)"; exit 1; }
+
+for ARGUMENT in "$@"
+do
+    KEY=$(echo "$ARGUMENT" | cut -f1 -d=)
+    VALUE=$(echo "$ARGUMENT" | cut -f2 -d=)
+
+    case "$KEY" in
+            url)      URL="$VALUE" ;;
+            *)        URL="$KEY" 
+    esac
+done
+
+if ! [[ -x "$(command -v wget)" ]]; then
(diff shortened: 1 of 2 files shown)


scripts/hash-exec-copy.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+set -e
+
+[[ "$1" =~ https://raw.githubusercontent.com/makerdao/community/* ]] || { echo "Plese provide the exec copy link to hash (e.g. url=<https://raw.githubusercontent.com/makerdao/community/...>)"; exit 1; }
```
The-Arbiter, line 4:
```suggestion
[[ "$1" =~ https://raw.githubusercontent.com/makerdao/community/* ]] || { echo "Please provide the exec copy link to hash (e.g. url=<https://raw.githubusercontent.com/makerdao/community/...>)"; exit 1; }
```

The-Arbiter (approved):
LGTM, checked against this week's exec and it works fine.

scripts/hash-exec-copy.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
gbalabasquer, line 1: As a general comment I'd say that this script might only make sense in `spells-mainnet`. Not opposed to have it here either, but won't be used.

scripts/hash-exec-copy.sh
```diff
@@ -0,0 +1,21 @@
...
+    VALUE=$(echo "$ARGUMENT" | cut -f2 -d=)
+
+    case "$KEY" in
+            url)      URL="$VALUE" ;;
+            *)        URL="$KEY" 
+    esac
+done
+
+if ! [[ -x "$(command -v wget)" ]]; then
+    cast keccak -- "$(curl "$URL" -o - 2>/dev/null)"
```
gbalabasquer, line 18: I think we should also make sure `curl` is available, otherwise to throw an error saying that `curl` or `wget` should be installed.

scripts/hash-exec-copy.sh
naszam: yeah, opened here for discussion, once approved I could port it to `spells-mainnet` and close the Draft PR.

scripts/hash-exec-copy.sh
```diff
@@ -0,0 +1,21 @@
...
+for ARGUMENT in "$@"
+do
+    KEY=$(echo "$ARGUMENT" | cut -f1 -d=)
+    VALUE=$(echo "$ARGUMENT" | cut -f2 -d=)
+
+    case "$KEY" in
+            url)      URL="$VALUE" ;;
+            *)        URL="$KEY" 
+    esac
+done
```
gbalabasquer, line 15: What's the main reason of this block? To be used when the script is called directly without using `make`?
If that's the case, I think we should tweak it a bit more as, these work:
```
./scripts/hash-exec-copy.sh https://raw.githubusercontent.com/makerdao/community/master/governance/votes/Executive%20Vote%20-%20April%2029%2C%202022.md
./scripts/hash-exec-copy.sh url=https://raw.githubusercontent.com/makerdao/community/master/governance/votes/Executive%20Vote%20-%20April%2029%2C%202022.md
```

which is nice, however if you make a typo in the `url` variable name, like:
```
./scripts/hash-exec-copy.sh uri=https://raw.githubusercontent.com/makerdao/community/master/governance/votes/Executive%20Vote%20-%20April%2029%2C%202022.md
```

It will silently succeed with the hash of empty value.

scripts/hash-exec-copy.sh
gbalabasquer: Hmm if will be just for mainnet, I guess we should do the draft PR there. Anyway not a big deal, just wondering if we also want the script here.

naszam: yeah, we could keep parity for the script and have it here as well, as you think it's best.

scripts/hash-exec-copy.sh
naszam: in the current setup, `curl` is a fallback for `wget` in case it's not installed, mostly for mac users where I think `wget` is not installed by default.

scripts/hash-exec-copy.sh
naszam: yeah, I guess the same could be for `get-created-timestamp.sh` but in this case it will actually cause an unexpected behaviour, nice catch!

scripts/hash-exec-copy.sh
naszam: it should be addressed in 89705da

scripts/hash-exec-copy.sh
naszam: it should be addressed in 89705da

scripts/hash-exec-copy.sh
gbalabasquer: Yeah fine for me. Let's keep it here as well.

scripts/hash-exec-copy.sh
gbalabasquer: Looks good now!

scripts/hash-exec-copy.sh
gbalabasquer: So now this doesn't work anymore:
```
./scripts/hash-exec-copy.sh https://raw.githubusercontent.com/makerdao/community/master/governance/votes/Executive%20Vote%20-%20April%2029%2C%202022.md
```

Which is better than before, maybe not the optimal.
In any case what I'd definitely add is a message error.

naszam: yeah, now it strictly enforce the `url` key I guess, I'll see what If there is a better approach.

naszam: this may fix it d4df582

gbalabasquer: So IMO the best it is to make it work with `url` and without it and fail with an error message with a wrong key.
Second best option is what we have now but adding the missing error message.
Anyway don't waste much time, we will be using it with `make` so to be honest will work the same way.

gbalabasquer: > this might fix it [d4df582](https://github.com/makerdao/spells-goerli/commit/d4df58271dd42f15960af07a977a6bfde4075d2b)

So this now allows to use any key that will work. And it is a bit redundant to do:
```
case "$KEY" in
            url)      URL="$VALUE" ;;
            *)        URL="$VALUE"
    esac
```
If we want this behaviour, you just remove the `case` and do plain `URL="$VALUE"`

naszam: yeah, not sure having both interoperability would make sense here

naszam: but maybe better removing the `case` as you suggested for this script as only one key is required

naszam: it should be done in dfcadb5

scripts/hash-exec-copy.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+set -e
+
+[[ "$1" =~ https://raw.githubusercontent.com/makerdao/community/* ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/...)"; exit 1; }
```
talbaneth, line 4: I like this validation, was about to propose it before I saw this line (as a common mistake is to use the non raw permalink path like https://github.com/makerdao/community/blob/master/governance/votes/Executive vote - October 5%2C 2022.md)

Consider maybe even doing `https://raw.githubusercontent.com/makerdao/community/*/governance/votes/*.md`

naszam: done in f53f8b9

scripts/hash-exec-copy.sh
```diff
@@ -0,0 +1,18 @@
...
+
+for ARGUMENT in "$@"
```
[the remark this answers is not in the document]
naszam: it should be addressed in f53f8b9

scripts/hash-exec-copy.sh
```diff
@@ -1,12 +1,9 @@
...
 set -e
 
-[[ "$1" =~ https://raw.githubusercontent.com/makerdao/community/* ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/...)"; exit 1; }
+URL=$(echo "$1" | cut -f2 -d=)
 
-for ARGUMENT in "$@"
-do
-    URL=$(echo "$ARGUMENT" | cut -f2 -d=)
-done
+[[ "$URL" == https://raw.githubusercontent.com/makerdao/community/*/governance/votes/*.md ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/<commit>/governance/votes/*.md)"; exit 1; }
```
gbalabasquer, line 6: Very cool that a regular expression check works so easily.

scripts/hash-exec-copy.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -e
+
+URL=$(echo "$1" | cut -f2 -d=)
+
+[[ "$URL" == https://raw.githubusercontent.com/makerdao/community/*/governance/votes/*.md ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/<commit>/governance/votes/*.md)"; exit 1; }
```
gbalabasquer, line 6:
```suggestion
[[ "$URL" == https://raw.githubusercontent.com/makerdao/community/*/governance/votes/*.md ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/<commit>/governance/votes/<file name>.md)"; exit 1; }
```

scripts/hash-exec-copy.sh
```diff
@@ -1,12 +1,9 @@
 #!/usr/bin/env bash
 set -e
 
-[[ "$1" =~ https://raw.githubusercontent.com/makerdao/community/* ]] || { echo "Please provide the correct exec copy link to hash (e.g. url=https://raw.githubusercontent.com/makerdao/community/...)"; exit 1; }
+URL=$(echo "$1" | cut -f2 -d=)
```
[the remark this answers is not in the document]
naszam: it should remove the key (`-f2`) with delimiter `=` (see `man cut`)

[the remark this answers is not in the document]
gbalabasquer: Ahh right, got it!

gbalabasquer (approved):
Looks good to be merged and ported to mainnet.

scripts/hash-exec-copy.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -e
+
+URL=$(echo "$1" | cut -f2 -d=)
```
talbaneth, line 4: I think it would be better to just take the variable as is, and remove the 2nd option of taking anything that comes after `=`. As it is now, you can run this script with `nonesense=<url>` and it would still work.

So in general we should either parse an exact field name, or not support giving the field name at all. For the case where there is only one input variable I see we mostly use the second option in the existing scripts (which is simpler).

talbaneth: see for example `archive-dssspell.sh`.

naszam: yes, I agree, for one arg is better the simple pattern, don in 748dcb

talbaneth (approved):
LGTM

The-Arbiter:
Agree with Gonza here, maybe a mainnet only change but good to have it across both I guess!

Looks good and thanks for this! :)

The-Arbiter (approved):

gbalabasquer (approved):
LGTM. The only thing I might suggest is to add a detection for the empty hash.
It would be nice to detect that and through an error. Let's say you pass a link that matches the regular expression but has an error which gives you a 404. Getting the hash of empty has been a pretty common error, in general just because we missed `wget` but can also happen due the reason presented above, or maybe a temporary not availability of the file.
Anyway writer and reviewer should always verify with the web functionality (copying and pasting).

naszam:
> LGTM. The only thing I might suggest is to add a detection for the empty hash. It would be nice to detect that and throw an error. Let's say you pass a link that matches the regular expression but has an error which gives you a 404. Getting the hash of empty has been a pretty common error, in general just because we missed `wget` but can also happen due the reason presented above, or maybe a temporary not availability of the file. Writer and reviewer should always verify with the web functionality (copying and pasting) but this extra check might be useful.

going to address the suggestion on a separate PR for discussion, as it might require some refactoring

PR #118, opened 2022-10-13
Add `ETHERSCAN_API_KEY` check in `deploy.sh`

naszam:
- Add `ETHERSCAN_API_KEY` check in `deploy.sh`

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -2,7 +2,7 @@
 set -e
 
 [[ "$(cast chain --rpc-url="$ETH_RPC_URL")" == "goerli" ]] || { echo "Please set a Goerli ETH_RPC_URL"; exit 1; }
-[[ "$ETHERSCAN_API_KEY" ]] || { echo "Please set a ETHERSCAN_API_KEY"; exit 1; }
+[[ "$ETHERSCAN_API_KEY" ]] || { echo "Please set ETHERSCAN_API_KEY"; exit 1; }
 
 make && \
   dapp create DssSpell | \
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -2,6 +2,7 @@
 set -e
 
 [[ "$(cast chain --rpc-url="$ETH_RPC_URL")" == "goerli" ]] || { echo "Please set a Goerli ETH_RPC_URL"; exit 1; }
+[[ "$ETHERSCAN_API_KEY" ]] || { echo "Please set a ETHERSCAN_API_KEY"; exit 1; }
 
 make && \
   dapp create DssSpell | \


talbaneth (dismissed):
LGTM

scripts/deploy.sh
```diff
@@ -2,6 +2,7 @@
 set -e
 
 [[ "$(cast chain --rpc-url="$ETH_RPC_URL")" == "goerli" ]] || { echo "Please set a Goerli ETH_RPC_URL"; exit 1; }
+[[ "$ETHERSCAN_API_KEY" ]] || { echo "Please set a ETHERSCAN_API_KEY"; exit 1; }
```
The-Arbiter, line 5: To be honest we have a bunch of env that we can enforce in this stage. This PR is for ES key check but we also have other stuff we should check for IMO in order to ensure it doesn't run when the env hasn't been configured. I also feel like echoing the values for gas limit and price might be a good idea.

```suggestion
[[ "$ETHERSCAN_API_KEY" ]] || { echo "Please set ETHERSCAN_API_KEY"; exit 1; }
[[ "$ETH_FROM" ]] || { echo "Please set ETH_FROM value (the address that the transaction is being sent from)"; exit 1; }
[[ "$ETH_KEYSTORE" ]] || { echo "Please set ETH_KEYSTORE (the keystore file location for ethsign to use)"; exit 1; }
[[ "$ETH_PASSWORD" ]] || { echo "Please set ETH_PASSWORD (the password.txt file location for ethsign to use)"; exit 1; }
[[ "$ETH_GAS" ]] || { echo "Please set ETH_GAS (gas limit)"; exit 1; }
[[ "$ETH_GAS_PRICE" ]] || { echo "Please set a ETH_GAS_PRICE (base gas price)"; exit 1; }
[[ "$ETH_PRIO_FEE" ]] || { echo "Please set a ETH_PRIO_FEE (priority gas price)"; exit 1; }
```

We also have future options here including [automatic gas price determination](https://book.getfoundry.sh/reference/cast/cast-gas-price?highlight=gas%20price#cast-gas-price) using cast or `seth gas-price` with dapptools, which can be used to save us on spell deployment costs.

The-Arbiter (approved):
lgtm

scripts/deploy.sh
naszam: I think local setup for deployment should be already documented in the readme, that could be improved as well.
Regarding the deployer, we could consider adding a check for the current deployer.
Let's see what is the general consensus and it could be addressed in a separate PR eventually, for discussion.
I've cleanup the error message in cea6cdd

talbaneth (approved):
LGTM

scripts/deploy.sh
gbalabasquer: I think I agree that we could check all the necessary variables. I haven't been personally using the script for deploying, but if you do it I see the added value to not let you go without some of them being set up.
Maybe I'd argue for `ETH_GAS` to be automatically set up running the estimation first and maybe with a threshold where if it goes above X it warns the user to continue.

gbalabasquer: The only thing we should make sure that if the script is converting to wei, to not do it twice if the user makesa mistake and actually sends the wei number.

The-Arbiter:
Just saw Gonza's replies. I agree that we should implement these changes however maybe we should have it as a separate PR to keep this first PR consistent across mainnet and goerli, then PR in the larger changes separately?

PR #132, opened 2022-11-04
Multichain spells

iamchrissmith:
# Description

This is the initial base PR for adding multichain support to the spells repo. It updates the forge test script and adds necessary config files to be able to **optionally** connect to optimism or arbitrum domains.  This will allow future development to add tests for specific values on the L2s or testing the effects of cross chain spells.

To connect to optimism or arbitrum when running tests you should export `OPT_RPC_URL` or `ARB_RPC_URL` along with your goerli `ETH_RPC_URL`.  Forge will load Goerli as a fork from your `ETH_RPC_URL`.  If you pass `arb-block` or `opt-block` into your `make test-forge` command (i.e. `make test-forge block=<GOERLI-BLOCK> opt-block=<OPTIMISM-BLOCK> arb-block=<ARBITRUM-BLOCK>`) your forks will be rolled to those specific blocks for testing the state/effects at that point in time.

## Note: this is a breaking change for `dapp` 

As we have to use cheatcodes (i.e. `vm.makePersistent`) that will not work for our `make test` command. As such this should probably go in after https://github.com/makerdao/spells-goerli/pull/130

### DIFF
--- a/config/domains.json
+++ b/config/domains.json
@@ -1,0 +1,17 @@
+{
+    "domains": {
+        "root": {
+            "rpc": "ETH_RPC_URL",
+            "chainlog": "0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F",
+            "block": "ROOT_BLOCK"
+        },
+        "optimism": {
+            "rpc": "OPT_RPC_URL",
+            "block": "OPT_BLOCK"
+        },
+        "arbitrum": {
+            "rpc": "ARB_RPC_URL",
+            "block": "ARB_BLOCK"
+        }
+    }
+}
--- a/config/domains.json
+++ b/config/domains.json
@@ -1,9 +1,7 @@
 {
     "domains": {
         "root": {
-            "rpc": "ETH_RPC_URL",
-            "chainlog": "0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F",
-            "block": "ROOT_BLOCK"
+            "chainlog": "0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F"
         },
         "optimism": {
             "rpc": "OPT_RPC_URL",
--- a/src/Goerli-DssSpell.t.base.sol
+++ b/src/Goerli-DssSpell.t.base.sol
@@ -18,7 +18,7 @@ pragma solidity 0.6.12;
 pragma experimental ABIEncoderV2;
 
 import "ds-math/math.sol";
-import "ds-test/test.sol";
+import "forge-std/Test.sol";
 import "dss-interfaces/Interfaces.sol";
 
 import "./test/rates.sol";
@@ -139,7 +139,7 @@ interface ArbitrumTeleportBridgeLike is TeleportBridgeLike {
     function inbox() external view returns (address);
 }
 
-contract GoerliDssSpellTestBase is Config, DSTest, DSMath {
+contract GoerliDssSpellTestBase is Config, Test, DSMath {
     Hevm hevm;
 
     Rates         rates = new Rates();
--- a/src/Goerli-DssSpell.t.base.sol
+++ b/src/Goerli-DssSpell.t.base.sol
@@ -289,12 +289,12 @@ contract GoerliDssSpellTestBase is Config, Test, DSMath {
             }
         }
     }
+    event Log(string, string);
+    event Log(string, uint256);
 
     function stringConcat(string memory a, string memory b) pure internal returns(string memory) {
         return string(abi.encodePacked(a, b));
     }
-
-    event Log(string, string);
 
     function readInput(string memory input) internal returns (string memory) {
         string memory root = vm.projectRoot();
@@ -303,15 +303,35 @@ contract GoerliDssSpellTestBase is Config, Test, DSMath {
         return vm.readFile(path);
     }
 
-    event Log(string, uint256);
-
-    function setUp() public {
+    function setUpDomains() public {
         vm.makePersistent(address(addr));
         config = readInput("domains");
         rootDomain = new Domain(vm, config, "root");
-        optimismDomain = new Domain(vm, config, "optimism");
-        rootDomain.selectFork();
+        // This should be set from the forge test script
+        rootDomain.loadFork(vm.activeFork());
 
+        // Load optimism from config
+        optimismDomain = new Domain(vm, config, "optimism");
+        optimismDomain.loadConfig();
+        assertEq(vm.activeFork(), 0);
+        assertEq(block.number, 7838100);
+
+        if (optimismDomain.live() == 1) {
+            optimismDomain.selectFork();
+            assertEq(vm.activeFork(), 1);
+            assertEq(block.number, 2437710, "wrong optimism block");
+            optimismDomain.rollFork(2437710 + 10);
+            assertEq(block.number, 2437720, "wrong optimism block");
+
+            rootDomain.selectFork();
+            assertEq(vm.activeFork(), 0);
+            assertEq(block.number, 7838100);
+        }
+    }
+
+
+    function setUp() public {
+        setUpDomains();
         setValues(address(chief));
 
         spellValues.deployed_spell_created = spellValues.deployed_spell != address(0) ? spellValues.deployed_spell_created : block.timestamp;
--- a/src/Goerli-DssSpell.t.base.sol
+++ b/src/Goerli-DssSpell.t.base.sol
@@ -311,7 +311,6 @@ contract GoerliDssSpellTestBase is Config, Test, DSMath {
         rootDomain = new Domain(vm, config, "root");
         optimismDomain = new Domain(vm, config, "optimism");
         rootDomain.selectFork();
-        rootDomain.rollFork();
 
         setValues(address(chief));
 
--- a/src/Goerli-DssSpell.t.base.sol
+++ b/src/Goerli-DssSpell.t.base.sol
@@ -18,13 +18,16 @@ pragma solidity 0.6.12;
 pragma experimental ABIEncoderV2;
 
 import "ds-math/math.sol";
-import "forge-std/Test.sol";
+import {Vm} from "forge-std/Vm.sol";
+import {Test} from "forge-std/Test.sol";
 import "dss-interfaces/Interfaces.sol";
 
 import "./test/rates.sol";
 import "./test/addresses_goerli.sol";
 import "./test/addresses_deployers.sol";
 import "./test/config.sol";
+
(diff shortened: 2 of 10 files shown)


config/domains.json
```diff
@@ -0,0 +1,15 @@
+{
```
hexonaut, line 1: Foundry convention is converging on putting this in `script/input/{chainId}/{name}.json` so `script/input/5/config.json` in this case.

src/Goerli-DssSpell.t.base.sol
```diff
@@ -287,21 +284,54 @@ contract GoerliDssSpellTestBase is Config, DSTest, DSMath {
...
-    function setUp() public {
-        hevm = Hevm(address(CHEAT_CODE));
+    function stringConcat(string memory a, string memory b) pure internal returns(string memory) {
+        return string(abi.encodePacked(a, b));
+    }
 
+    function readInput(string memory input) internal returns (string memory) {
+        string memory root = vm.projectRoot();
+        string memory path = stringConcat(stringConcat(root, "/config/"), stringConcat(input, ".json"));
+        emit Log("path", path);
```
hexonaut, line 302: Was this accidentally left in for debug?

src/test/domains/Domain.sol
```diff
@@ -0,0 +1,93 @@
...
+    using stdJson for string;
+
+    string public config;
+    string public name;
+    Vm public vm;
+    uint256 public forkId;
+    uint256 public live = 0;
+
+    event Log(string, string);
+    event Log(string, uint256);
```
hexonaut, line 33: I think these were left in for debug by accident.

src/Goerli-DssSpell.t.base.sol
```diff
@@ -139,8 +132,12 @@ interface ArbitrumTeleportBridgeLike is TeleportBridgeLike {
...
 }
 
-contract GoerliDssSpellTestBase is Config, DSTest, DSMath {
-    Hevm hevm;
+contract GoerliDssSpellTestBase is Config, Test, DSMath {
+
+    string config;
+    Domain rootDomain;
+    Domain optimismDomain;
+    Domain arbitrumDomain;
```
hexonaut, line 140: We probably want to use the `ArbitrumDomain` and `OptimismDomain` types here so that we can replay txs from L1 to L2. IE the `XXXDomain.sendToHost(...)` call for full integration testing. This will help us to verify that the L2 part of the spell has been queued up properly in the CTC.

hexonaut (changes requested):
This looks like a good start. I like the additions of the block selection to the domain. Maybe we can backport that to `dss-test`.

I'm wondering if we want to actually import `dss-test` instead of copying the `Domain` in here. Everything currently would be basically the same except you load in the `XXXDomain` stuff from `dss-test`.

src/Goerli-DssSpell.t.base.sol
iamchrissmith: My thought here was that since we didn't have a specific use for this right now in the repo, leave it basic and then add these when we have the need.  That being said, if you feel strongly i can put the code in since it would mostly be copy/paste from your work

src/test/domains/Domain.sol
iamchrissmith: https://github.com/makerdao/spells-goerli/pull/132/commits/28d3dc4e0d8870cd0b47213b3a1b24014d9f2035

src/Goerli-DssSpell.t.base.sol
iamchrissmith: https://github.com/makerdao/spells-goerli/pull/132/commits/28d3dc4e0d8870cd0b47213b3a1b24014d9f2035

config/domains.json
hexonaut: The chain id is the root chain (Mainnet or Goerli). I'm following some conventions from other Foundry repos as these standards get built.

Maybe since this repo is entirely Goerli an exception can be made, but I don't see any issue with keeping the convention either.

src/Goerli-DssSpell.t.base.sol
hexonaut: Well the whole point of this addition is we can do integration tests for Optimism and Arbitrum. :)

Also, as I said in the primary comment I think you can just import them from `dss-test` instead of copying in. If you need code changes (such as selecting the block) it should be merged upstream to `dss-test` if it's missing functionality.

gbalabasquer: I'd also think we should be including `dss-test`. This repo should be just another one that takes usage of that one.
We are still on development phases of `dss-test` so whatever is needed for this repo we should be able to make sure is compatible.

gbalabasquer:
Btw how are we thinking to structure the directories for the actual L2 spell. To be honest I'm not very aware how the Governance relay works, so not sure which parts it will have. Will there be a Spell, spell action, ExecLib as well?

hexonaut:
> Btw how are we thinking to structure the directories for the actual L2 spell. To be honest I'm not very aware how the Governance relay works, so not sure which parts it will have. Will there be a Spell, spell action, ExecLib as well?

L2 relay is more straight forward than pause proxy setup as it is just a delegatecall with no delay. You deploy a spell (or really any piece of code that can be executed) then instruct the L1 side of the relay to call the destination function on the L2 previously deployed contract. I think we can just copy the spell structure almost entirely. DssExecLib needs to be deployed on L2 for support.

gbalabasquer:
> > Btw how are we thinking to structure the directories for the actual L2 spell. To be honest I'm not very aware how the Governance relay works, so not sure which parts it will have. Will there be a Spell, spell action, ExecLib as well?
> 
> L2 relay is more straight forward than pause proxy setup as it is just a delegatecall with no delay. You deploy a spell (or really any piece of code that can be executed) then instruct the L1 side of the relay to call the destination function on the L2 previously deployed contract. I think we can just copy the spell structure almost entirely. DssExecLib needs to be deployed on L2 for support.

Ok, sounds good, so more than a `DssSpell` for L2s, we need just a `DssSpellAction` with its own `DssExecLib` deployed in L2 (this will actually be more necessary when MCD is deployed). Still, we need to figure it out the correct way to build the structure in this repo for weeks that require the L2 action.

gbalabasquer:
I think a good spell exercise (and example) would be adding the chainlog to the 2 Arbitrum and Optimism. We can just create the chainlogs, assign the ownership to the gov relay and create the L1 spell with the 3 L2s actions that will set up the keys.

src/Goerli-DssSpell.t.base.sol
iamchrissmith: so the problem with `dss-test` is that it is using `>= 0.8.0` which we have so far wanted to keep `spells-*` and `dss-exec-lib` at `0.6.12` for safety reasons.  If we want to use `dss-test` then we need to upgrade those and that should be a broader team discussion/decision

hexonaut: Ahh yes good point. Maybe it's time to look at upgrading the spells repo version, but that's definitely a bigger change.

gbalabasquer: I don't really think there is a safety reason to not upgrade the spells to 0.8.x, more if we are doing it without optimizations.
Regarding `DssExecLib`, remember we can keep the deployed library with solc 0.6.12 and just use the new solidity version when inheriting `DssExec` and `DssAction`.

config/domains.json
iamchrissmith: https://github.com/makerdao/spells-goerli/pull/132/commits/e3a857bce192819982700b9e3b402a358fa08fb9

src/Goerli-DssSpell.t.base.sol
iamchrissmith: > Regarding `DssExecLib`, remember we can keep the deployed library with solc 0.6.12 and just use the new solidity version when inheriting `DssExec` and `DssAction`.

You're saying we don't have to update the deployed contract, but we do need to still update the repo, correct?

src/Goerli-DssSpell.t.base.sol
```diff
@@ -296,13 +296,13 @@ contract GoerliDssSpellTestBase is Config, Test, DSMath {
 
     function readInput(string memory input) internal returns (string memory) {
         string memory root = vm.projectRoot();
-        string memory path = stringConcat(stringConcat(root, "/config/"), stringConcat(input, ".json"));
+        string memory path = stringConcat(stringConcat(root, "/scripts/input/5/"), stringConcat(input, ".json"));
         return vm.readFile(path);
     }
```
hexonaut, line 301: Can you use this code?

```
function readInput(string memory input) internal returns (string memory) {
	string memory root = vm.projectRoot();
	string memory chainInputFolder = string.concat("/script/input/", vm.toString(block.chainid), "/");
	return vm.readFile(string.concat(root, chainInputFolder, string.concat(input, ".json")));
}
```

src/test/domains/Domain.sol
```diff
@@ -0,0 +1,90 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+// Copyright (C) 2022 Dai Foundation
```
naszam, line 2:
```suggestion
// SPDX-FileCopyrightText: © 2022 Dai Foundation <www.daifoundation.org>
// SPDX-License-Identifier: AGPL-3.0-or-later
```

src/Goerli-DssSpell.t.base.sol
iamchrissmith: yeah so unfortunately both `block.chainId` and `string.concat` are a solidity 0.8 things. so the 0.6.12 code needs to become:

```solidity
function getChainID() external view returns (uint256) {
    uint256 id;
    assembly {
        id := chainid()
    }
    return id;
}

function readInput(string memory input) internal returns (string memory) {
    string memory root = vm.projectRoot();
    string memory chainInputFolder = stringConcat(stringConcat("/scripts/input/", vm.toString(getChainId())),"/");
    string memory path = stringConcat(stringConcat(root, chainInputFolder), stringConcat(input, ".json"));
    return vm.readFile(path);
}
```

I'm not sure that is worth it to avoid having to hardcode 5 for goerli and 1 for mainnet.  Do you think it is worth that?

For "when we update', we're going to set a meeting on this, but from the planning call, it sounded like it might not have support.  I think it requires a lot of effort and there are still optimizer bugs being discovered in 0.8.x.  One possible option would be to move multichain stuff to its own repo and import that into spells and xdomain, but then this code would still be 0.6.12 for spells

hexonaut: Sure. Not a big deal if you want to leave it like this I guess.

PR #147, opened 2023-01-04
Remove template directory + Standardize use of pragma

gbalabasquer:
Removing the template directory as not really used.


megozu (approved):
LGTM

megozu:
should we bump `forge-std` as well to `v1.2.0`?

gbalabasquer:
> should we bump `forge-std` as well to `v1.2.0`?

Let's leave that for another moment.

iamchrissmith (approved):
lgtm

PR #150, opened 2023-01-11
Add dss-test

iamchrissmith:
# Description
Switch from forge-std/Test.sol to dss-test/DSSTest.sol

Draft until we can merge and update for:
- [x] TODO: Update DssTest setBalance to handle Vyper tokens and switch from `_giveTokens` to `setBalance` (started: https://github.com/makerdao/dss-test/pull/21)

### DIFF
--- a/src/DssSpell.t.base.sol
+++ b/src/DssSpell.t.base.sol
@@ -641,64 +641,9 @@ contract DssSpellTestBase is Config, DSSTest {
     }
 
     function _giveTokens(address token, uint256 amount) internal {
-        // Edge case - balance is already set for some reason
-        if (GemAbstract(token).balanceOf(address(this)) == amount) return;
-
-        // Scan the storage for the balance storage slot
-        for (uint256 i = 0; i < 200; i++) {
-            // Solidity-style storage layout for maps
-            {
-                bytes32 prevValue = vm.load(
-                    address(token),
-                    keccak256(abi.encode(address(this), uint256(i)))
-                );
-
-                vm.store(
-                    address(token),
-                    keccak256(abi.encode(address(this), uint256(i))),
-                    bytes32(amount)
-                );
-                if (GemAbstract(token).balanceOf(address(this)) == amount) {
-                    // Found it
-                    return;
-                } else {
-                    // Keep going after restoring the original value
-                    vm.store(
-                        address(token),
-                        keccak256(abi.encode(address(this), uint256(i))),
-                        prevValue
-                    );
-                }
-            }
-
-            // Vyper-style storage layout for maps
-            {
-                bytes32 prevValue = vm.load(
-                    address(token),
-                    keccak256(abi.encode(uint256(i), address(this)))
-                );
-
-                vm.store(
-                    address(token),
-                    keccak256(abi.encode(uint256(i), address(this))),
-                    bytes32(amount)
-                );
-                if (GemAbstract(token).balanceOf(address(this)) == amount) {
-                    // Found it
-                    return;
-                } else {
-                    // Keep going after restoring the original value
-                    vm.store(
-                        address(token),
-                        keccak256(abi.encode(uint256(i), address(this))),
-                        prevValue
-                    );
-                }
-            }
-        }
-
-        // We have failed if we reach here
-        assertTrue(false, "TestError/GiveTokens-slot-not-found");
+        // This passes through to the GodMode function. Because we have another logic branch
+        // in spells-mainnet, we're keeping this wrapper here for consistency
+        GodMode.setBalance(token, address(this), amount);
     }
 
     function _checkIlkIntegration(
--- a/src/DssSpell.t.base.sol
+++ b/src/DssSpell.t.base.sol
@@ -16,7 +16,6 @@
 
 pragma solidity 0.8.16;
 
-import {DSSTest} from "dss-test/DSSTest.sol";
 import "dss-interfaces/Interfaces.sol";
 
 import "./test/rates.sol";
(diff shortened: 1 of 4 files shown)


iamchrissmith:
Switching the inheritance to go through `Config` seemed to be the best way to handle the conflicting math constant declarations.  lmk if there is a different preferred way.

gbalabasquer:
> Switching the inheritance to go through `Config` seemed to be the best way to handle the conflicting math constant declarations. lmk if there is a different preferred way.

Hmmm is a bit weird `Config` inherits from `DssTest`, what would happen if you declare the unit constants in `Config` as `private` which should provoke them are inherited.

hexonaut (commented):
Changes so far look good. I would leave the `dss-test` `GodMode.setBalance()` to just do the Vyper and Solidity, and you can replace the code after the GUSD check (leaving the bespoke GUSD code in-tact).

Re: config inheritance - you can set the constants to private and it will compile. I would leave the inheritance of DSSTest in the base contract instead of config and provide private `MILLION`, etc in `config.sol`.

gbalabasquer:
Yeah the GUSD solution seems fine for me. I mean that will be for mainnet, here we could use setBalance directly, but to keep consistency between the two repos I would still keep here giveTokens and call setBalance inside.
Regarding the `private` thing, great you confirmed it fixes the problem.

hexonaut (dismissed):
LGTM

src/DssSpell.t.base.sol
```diff
@@ -221,11 +213,11 @@ contract DssSpellTestBase is Config, Test {
     }
 
     function _concat(string memory a, string memory b) internal pure returns (string memory) {
-        return string(abi.encodePacked(a, b));
+        return string.concat(a, b);
     }
 
     function _concat(string memory a, bytes32 b) internal pure returns (string memory) {
-        return string(abi.encodePacked(a, _bytes32ToStr(b)));
+        return string.concat(a, _bytes32ToStr(b));
```
The-Arbiter, line 220: Just going to dig in here, not raising an issue directly

So abi.encodePacked will pack the values (remove leading zeroes) whereas I don't think string.concat does

See https://docs.soliditylang.org/en/v0.8.17/units-and-global-variables.html?highlight=string%20concat#members-of-string

And https://docs.soliditylang.org/en/v0.8.17/units-and-global-variables.html?highlight=abi.encodepacked#abi-encoding-and-decoding-functions
Image

I don't think there are going to be any issues here since string.concat gets given two strings (we cast using `string()` in `_bytes32ToStr`) but I wanted to confirm that this has been considered and the outputs are never different (I don't think they ever are).

src/test/config.sol
```diff
@@ -82,10 +82,10 @@ contract Config {
         bool    offboarding;
     }
 
-    uint256 constant HUNDRED    = 10 ** 2;
-    uint256 constant THOUSAND   = 10 ** 3;
-    uint256 constant MILLION    = 10 ** 6;
-    uint256 constant BILLION    = 10 ** 9;
+    uint256 constant         HUNDRED  = 10 ** 2;
```
The-Arbiter, line 85:
```suggestion
    uint256 constant private HUNDRED  = 10 ** 2;
```

Or is there a reason for no vis declaration (can use public instead?)

I see [it isn't in dss-test](https://github.com/makerdao/dss-test/blob/52cdc7cca6093e59ec4319b8647c32db72a4f24e/src/DSSTest.sol#L47) so do we want to formalise HUNDRED?

The-Arbiter: HUNDRED is used once for WBTC-C `dust` (Can just be 7_500 or 75_00 instead). IMO reliance on less constants / vars is useful?

The-Arbiter (approved):
LGTM. One potential typo fix and one concern about equivalence to the existing setup.

src/test/config.sol
The-Arbiter, line 85: Optional reduction of constant reliance
```suggestion
```

Also requires `dust:         75 * HUNDRED,` in WBTC (Line 384) changing to `dust:         75_00,`

Which IMO is the way to go

The-Arbiter (commented):
Optional reduction of constant reliance (HUNDRED)

src/test/config.sol
gbalabasquer: IMO `HUNDRED` doesn't really make much sense, so I would vote to remove it as well.

hexonaut: Ya agreed to remove.

iamchrissmith: removed it

src/DssSpell.t.base.sol
iamchrissmith: good catch. It seems to me they would be the same since as you say we're giving it 2 strings.  Can you come up with a test that would prove this?

hexonaut (dismissed):
LGTM. lol these are some weird use cases for `HUNDRED`...

gbalabasquer (dismissed):
LGTM

src/DssSpell.t.base.sol
The-Arbiter: I agree that it would be the same.

We *could* in theory test this by generating random bytes strings (with leading zeroes) that get cast to string and then comparing the output, but (1) generating these strings would be very painful (even in Foundry) and (2) it doesn't feel like it fits within scope of what is contained in this repo.

I think if we want to do a side investigation on this, that would be a good idea, but it is probably not suited here. 

Again I *am basically certain* that our outputs should be the same but it's good to consider edge cases.

The-Arbiter: https://github.com/The-Arbiter/to_string_or_not_to_string

The-Arbiter (approved):
[I looked into the string concerns and it looks fine](https://github.com/The-Arbiter/to_string_or_not_to_string) but I didn't exhaustively test how `string()` works under the hood.

LGTM

iamchrissmith:
sorry @gbalabasquer , @hexonaut , and @The-Arbiter changing the base dismissed your :heavy_check_mark: s, so need new ones

hexonaut (approved):

gbalabasquer (approved):

PR #153, opened 2023-01-18
Improve CI results by introducing rollFork

iamchrissmith:
Adding tracking for the block number for the deployed spell allows us to rollFork to that block so our test will continue to pass even after the spell has been scheduled/cast. This should give us more confidence in our CI results as there aren't times that it "should fail".


gbalabasquer (dismissed):
LGTM

naszam (dismissed):
- Checked and discussed cheatcode changes with @iamchrissmith 
- Left a nit for consistency with the other comments

LGTM :+1:

gbalabasquer (approved):

naszam (dismissed):

src/test/config.sol
```diff
@@ -96,6 +97,7 @@ contract Config {
         spellValues = SpellValues({
             deployed_spell:                 address(0xF39BBaaAf13b84e535Ca85E6724054a510457812),     // populate with deployed spell if deployed
             deployed_spell_created:         1674571680,              // use `./scripts/get-created-timestamp.sh <deployment-tx>`
+            deployed_spell_block:           8368491,        // populate with the block where the spell was deployed
```
brianmcmichael, line 100: Is there a way to derive this from the timestamp or vice-versa, that way we don't need to set the additional variable each time.

gbalabasquer: I don't see a clean way to do so. The only way it comes to my mind is in the test script, get the bash to look for the block number in the solc file, call `cast block <number> timestamp` and then save that value in a temp file, which with forge we can import in our tests. But not very elegant solution.

iamchrissmith: I thought i could do this with `block.timestamp` but it doesn't look like I can without also rolling to the deploy block which would require an archive node, so i think we should keep this as-is for now

naszam: maybe a feat. request could be issued to foundry to consider extending the `rollFork` functionality using timestamp as input, instead of block number.

iamchrissmith: that would be nice, but since we already have a script to pull the timestamp and the contract address easily gives us the block number, i would say we not wait on this PR until we can populate one from the other without an archive node

naszam (approved):
LGTM (as first iteration, to be improved following developments of the tooling)

godsflaw (dismissed):

brianmcmichael (approved):

naszam (approved):
LGTMerge

brianmcmichael (approved):

PR #159, opened 2023-01-27
Disable Starknet l2 gateway

maciejka:
# Description

# Contribution Checklist

- [ ] Code approved
- [ ] Tests approved
- [ ] CI Tests pass

# Checklist

- [ ] Deploy spell to Goerli `ETH_GAS_LIMIT="XXX" ETH_GAS_PRICE="YYY" make deploy`
- [ ] Cast spell on Goerli `make spell="0x-deployed-spell-address" cast-spell`
- [ ] Run `make archive-spell` or `make date="YYYY-MM-DD" archive-spell` to make an archive directory and copy `DssSpell.sol`, `DssSpell.t.sol` and `DssSpell.t.base.sol`


src/DssSpell.sol
```diff
@@ -112,6 +125,17 @@ contract DssSpellAction is DssAction {
             ARB_GAS_PRICE_BID,
             ARB_MAX_SUBMISSION_COST
         );
+
+        // ------------------ Pause Starknet Goerli L2DaiTeleportGateway -----------------
+        // Forum: https://forum.makerdao.com/t/community-notice-pecu-to-redeploy-teleport-l2-gateways/19550
+        // L2 Spell to execute via STARKNET_GOV_RELAY:
+        // src: https://github.com/makerdao/starknet-spells-goerli/blob/b7ca995cf1d266aa2382d85e35a86b4fae52aa15/src/spell.cairo
+        // contract: https://testnet.starkscan.co/class/0x00a052591661d7e249b46a1084c63b14dae6aa8b1a56ab3f7df8c8add1c374b1#overview
+        StarknetGovRelayLike(STARKNET_GOV_RELAY).relay{value: STA_L1_CALL_VALUE}(STARKNET_L2_SPELL);
```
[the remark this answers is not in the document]
rockyfour: Mmm even if STARKNET_GOV_RELAY pre-receives ETH through `selfdestruct`, the actual [call](https://goerli.etherscan.io/address/0x8919aefA417745F22c6af5AD6550E83159a373F3#code#F1#L49) it makes to `starknet` is passing `msg.value`, which would be 0. 

So if `starknet` expects eth I think the options are:
1. Pre-fund the pause proxy using `selfdestruct` as already suggested (and seems like the current code supports).
2. Add an authed intermediate contract between the pause proxy and the gov relay and fund it normally.
3. Redeploy the governance relays on both sides, adding a default payable function on l1.

IMO option 3 is the least hacky and also aligns with the Arbitrum intended scheme.

src/DssSpell.sol
```diff
@@ -112,6 +125,17 @@ contract DssSpellAction is DssAction {
...
+
+        // disallow legacy bridge on escrow
+        StarknetEscrowLike(STARKNET_ESCROW).approve(DAI, STARKNET_DAI_BRIDGE_LEGACY, 0);
```
rockyfour, line 137: Can you please give some context to why we are doing this now.

src/DssSpell.sol
```diff
@@ -63,6 +67,11 @@ contract DssSpellAction is DssAction {
     uint256 public constant ARB_MAX_SUBMISSION_COST = 1e14;
     uint256 public constant ARB_L1_CALL_VALUE = ARB_MAX_SUBMISSION_COST + ARB_MAX_GAS * ARB_GAS_PRICE_BID;
 
+    // see: https://github.com/makerdao/starknet-spells-goerli/tree/teleport-spell#estimate-l1-l2-fee
+    uint256 public constant STA_FEE_ESTIMATION = 104825981483660;
```
rockyfour, line 71: Some questions:

1. If this value is not enough would it fail on l1 or just not be included on l2?

2. In case it is just not included in l2, can it be retried permissionlessly somehow with a larger gas fee? (in Optimism and Arbitrum it can in case it reverts on l2 for not enough gas).

3. It seems that we just use the estimated result, while in practice the spell will be executed more than 2 days later. For Abitrum there is a 20x factor for that now, do we need one here as well?

maciejka: 1. If there is not enough gas spell will 'fail' on l2, from l1 perspective l2 part of the spell will seem to not have been executed. 
2. Right now there is no permissionless mechanism to retry failed l1->l2 message. It would need to be built into the L1 gov relay using message cancelation. As there is a large upgrade coming in next few months to the way Starknet handles failed transactions we felt it was not worth the complication.
3. We need a safety buffer here as well. That is why we have STA_FEE_BUFFER. How the 20x was deceided?

src/DssSpell.sol
maciejka: It is part of the two step bridge upgrade process described here: https://github.com/makerdao/starknet-dai-bridge#upgrades. It was not executed earlier as there were a lot of other stuff going into the spells at the end of last year. For the same reasons L2 part of the spell includes also code the cleans up GovRelay upgrade that happend in October '22.

src/DssSpell.sol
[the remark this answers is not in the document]
maciejka: Redeployment has to go through current STARKNET_GOV_RELAY so it would need to use option 1 or 2 anyway. 
There is a large upgrade planned on `starknet` in next few months. It will make us to redeploy STARKNET_GOV_RELAY. It will be a good moment to upgrade it without any extra work. Until then I would use #1 or #2.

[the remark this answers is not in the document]
EvolveArt: Option 1 seems like the way to go for this one, as @maciejka mentionned we will have to rewrite it anyway in the short term future.

src/DssSpell.sol
maciejka: The problem with setting excess fee for l1->l2 calls in Starknet to acomodate gas price changes is that currently unused fees are not returned.

src/DssSpell.sol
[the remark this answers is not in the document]
rockyfour: We decided to go for option 1, which does not require code changes.
@olivdb I think this can be resolved.

godsflaw:
I ported these changes over in this commit: https://github.com/makerdao/spells-goerli/pull/160/commits/9b2b96bfcf857b06a87759e00ef19a381e839ec4

PR #163, opened 2023-01-31
Update dss test custom path

iamchrissmith:
Pulls in a change in dss-test that allows us to use the config file in dss-test instead of copying it here.


godsflaw (approved):
LGTM.  Can merge on 🟢 .

PR #164, opened 2023-02-01
Test that starknet spell is scheduled in core

brianmcmichael:
Tests that a starknet spell is scheduled in starknet core after cast

### DIFF
--- a/src/test/starknet.t.sol
+++ b/src/test/starknet.t.sol
@@ -177,7 +177,7 @@ contract StarknetTests is DssSpellTestBase, ConfigStarknet {
 
             bytes32 _message = _getL1ToL2MsgHash(addr.addr("STARKNET_GOV_RELAY"), starknetValues.l2_gov_relay, RELAY_SELECTOR, _payload, _nonce);
 
-            // Assert message is scheduled
+            // Assert message is scheduled, core returns 0 if not in message array
             assertTrue(core.l1ToL2Messages(_message) > 0, "StarknetTest/SpellNotQueued");
         }
     }
--- a/src/test/starknet.t.sol
+++ b/src/test/starknet.t.sol
@@ -169,6 +169,7 @@ contract StarknetTests is DssSpellTestBase, ConfigStarknet {
 
             // Nonce increments each message, back up one
             uint256 _nonce = core.l1ToL2MessageNonce() - 1;
+
             // Payload must be array
             uint256[] memory _payload = new uint256[](1);
             _payload[0] = uint256(_spell);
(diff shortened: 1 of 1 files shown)


src/test/starknet.t.sol
```diff
@@ -154,4 +161,48 @@ contract StarknetTests is DssSpellTestBase, ConfigStarknet {
 
         assertTrue(core.isNotFinalized());
     }
+
+    function _checkStarknetMessage(bytes32 _spell) internal {
+        StarknetCoreLike core = StarknetCoreLike(addr.addr("STARKNET_CORE"));
```
maciejka, line 166: Why not test the GovRelay itself? I mean instead of replicating `relay` method logic it can be called directly using forge prank mode.

src/test/starknet.t.sol
```diff
@@ -154,4 +161,48 @@ contract StarknetTests is DssSpellTestBase, ConfigStarknet {
...
+            _payload[0] = uint256(_spell);
+
+            // Hardcoded in L1 gov relay, not public
+            uint256 RELAY_SELECTOR = 300224956480472355485152391090755024345070441743081995053718200325371913697;
+
+            // Hash of message created by Starknet Core
+            bytes32 _message = _getL1ToL2MsgHash(addr.addr("STARKNET_GOV_RELAY"), starknetValues.l2_gov_relay, RELAY_SELECTOR, _payload, _nonce);
+
+            // Assert message is scheduled, core returns 0 if not in message array
+            assertTrue(core.l1ToL2Messages(_message) > 0, "StarknetTest/SpellNotQueued");
```
maciejka, line 184: StarknetCore contract emits `LogMessageToL2`. It is better object to test against than the content of `l1Tol2Messages` which should be considered an implementation detail.
The way messages are hashed might change in the future.

rockyfour: This is a good idea. Can use `vm.expectEmit` for that.

src/test/starknet.t.sol
brianmcmichael: This test happens after the spell is cast in the setup, so attempting to re-relay here will result in a different nonce and hash.

src/test/starknet.t.sol
brianmcmichael: `emit LogMessageToL2(msg.sender, toAddress, selector, payload, nonce, msg.value);`

This value is returned piecemeal and will still need to be re-hashed in the spell to reconstruct the message in the same manner as above. I don't think it gets us around any implementation-specific details, so the tests will need to be updated if this implementation changes the hash parameters in either case.

brianmcmichael: The fully hashed message is returned from the 

`StarkNetLike(starkNet).sendMessageToL2{value: msg.value}(l2GovernanceRelay, RELAY_SELECTOR, payload);` 

call. If anything, the next version of the Gov Relay should probably include an event that captures that message and emits the return from it.

src/test/starknet.t.sol
maciejka: You are right. I misread your test, I was under impression that it tries to send l1Tol2 message. 
Still, I think that it test internal details of Starknet core contract which is not necessary. Contract of `sendMessageToL2` method is that if it does not revert, message is queued.

src/test/starknet.t.sol
maciejka: Technically, Starknet sequencer initates l2 transactions based on `LogMessageToL2` events. `l1Tol2Messages` prevents it from including spoofed transactions that were not initiated from l1.

godsflaw:
I'm going to merge this now so that we can deploy soon.  I think if there are changes based on implementation these tests will break in the worst case.  We can just fix them if/when that day comes.

PR #166, opened 2023-02-11
Utilizes rollfork to obsolete deploy-info script

brianmcmichael:
# Description

Builds upon the rollFork innovation to use binary search for deployed block.

### DIFF
--- a/src/DssSpell.t.base.sol
+++ b/src/DssSpell.t.base.sol
@@ -266,22 +266,50 @@ contract DssSpellTestBase is Config, DssTest {
     function setUp() public {
         setValues(address(chief));
 
-        spellValues.deployed_spell_created = spellValues.deployed_spell != address(0) ? spellValues.deployed_spell_created : block.timestamp;
-        _castPreviousSpell();
-        spell = spellValues.deployed_spell != address(0) ?
-            DssSpell(spellValues.deployed_spell) : new DssSpell();
-
-        if (spellValues.deployed_spell_block != 0 && spell.eta() != 0) {
-            // if we have a deployed spell in the config
-            // we want to roll our fork to the block where it was deployed
-            // this means the test suite will continue to accurately pass/fail
-            // even if mainnet has already scheduled/cast the spell
-            vm.makePersistent(address(this));
-            vm.makePersistent(address(rates));
-            vm.makePersistent(address(addr));
-            vm.makePersistent(address(deployers));
-            vm.rollFork(spellValues.deployed_spell_block);
+        if (spellValues.deployed_spell != address(0)) {
+            spell = DssSpell(spellValues.deployed_spell);
+            if (spell.eta() != 0) {
+                // if we have a scheduled spell in the config
+                // we want to roll our fork to the block where it was deployed
+                // this means the test suite will continue to accurately pass/fail
+                // even if mainnet has already scheduled/cast the spell
+                vm.makePersistent(address(this));
+                vm.makePersistent(address(rates));
+                vm.makePersistent(address(addr));
+                vm.makePersistent(address(deployers));
+                _rollToDeployedSpellBlock(address(spell));
+            }
+        } else {
+            spell = new DssSpell();
         }
+
+        _castPreviousSpell();
+    }
+
+    function _rollToDeployedSpellBlock(address _spell) internal {
+        if (_spell == address(0)) { vm.rollFork(block.number); }
+
+        uint256 low = 5273074; // MCD_VAT Deployed Aug-06-2021 04:26:38 PM +UTC
+        uint256 high = block.number;
+        uint256 mid;
+
+        while (low != high) {
+            mid = (low + high) / 2;
+            vm.rollFork(mid);
+            if (_isContractDeployed(_spell)) {
+                high = mid - 1;
+            } else {
+                low = mid + 1;
+            }
+        }
+
+        return vm.rollFork(low + 1);
+    }
+
+    function _isContractDeployed(address _spell) internal view returns (bool) {
+        uint256 size;
+        assembly { size := extcodesize(_spell) }
+        return size > 0;
     }
 
     function _vote(address spell_) internal {
@@ -1446,11 +1474,7 @@ contract DssSpellTestBase is Config, DssTest {
         assertEq(_stringToBytes32(spell.description()),
                 _stringToBytes32(description), "TestError/spell-description");
 
-        if(address(spell) != address(spellValues.deployed_spell)) {
-            assertEq(spell.expiration(), block.timestamp + spellValues.expiration_threshold, "TestError/spell-expiration");
-        } else {
-            assertEq(spell.expiration(), spellValues.deployed_spell_created + spellValues.expiration_threshold, "TestError/spell-expiration");
-        }
+        assertEq(spell.expiration(), block.timestamp + spellValues.expiration_threshold, "TestError/spell-expiration");
 
         assertTrue(spell.officeHours() == spellValues.office_hours_enabled, "TestError/spell-office-hours");
 
--- a/src/test/config.sol
+++ b/src/test/config.sol
@@ -20,8 +20,6 @@ contract Config {
 
     struct SpellValues {
         address deployed_spell;
-        uint256 deployed_spell_created;
-        uint256 deployed_spell_block;
         address previous_spell;
         bool    office_hours_enabled;
         uint256 expiration_threshold;
@@ -96,8 +94,6 @@ contract Config {
         //
         spellValues = SpellValues({
             deployed_spell:                 address(0x89625bb73bA7Dccde985F690CEa2659525AB5b15),     // populate with deployed spell if deployed
-            deployed_spell_created:         1675789032,              // use `./scripts/get-created-timestamp.sh <deployment-tx>`
-            deployed_spell_block:           8451434,        // populate with the block where the spell was deployed
             previous_spell:                 address(0),     // supply if there is a need to test prior to its cast() function being called on-chain.
             office_hours_enabled:           false,          // true if officehours is expected to be enabled in the spell
             expiration_threshold:           30 days         // Amount of time before spell expires


src/DssSpell.t.base.sol
```diff
@@ -266,22 +266,53 @@ contract DssSpellTestBase is Config, DssTest {
...
+            mid = (low + high) / 2;
+            vm.rollFork(mid);
+            if (_isContractDeployed(_spell)) {
+                high = mid - 1;
+            } else {
+                low = mid + 1;
+            }
         }
+
+        vm.rollFork(low + 1);
```
naszam, line 309: Tested:
- `make test` PASS
- block [8451434](https://goerli.etherscan.io/block/8451434) deployed PASS
- block [8451989](https://goerli.etherscan.io/block/8451989) cast PASS
- block [8451982](https://goerli.etherscan.io/block/8451982) schedule, gets stuck

Then, playing around I found this pattern:

- 8451982, gets stuck `make test block=8451982`
- 8451983 pass
- 8451984 fail
- 8451985 stuck
- 8451986 pass
- 8451987 fail
- 8451988 stuck
- ...

naszam: Addressed in #167

naszam (approved):
LGTM

src/DssSpell.t.base.sol
brianmcmichael: Thank you!

src/DssSpell.t.base.sol
```diff
@@ -266,22 +266,53 @@ contract DssSpellTestBase is Config, DssTest {
...
-            vm.rollFork(spellValues.deployed_spell_block);
+    }
+
+    function _rollToDeployedSpellBlock(address _spell) internal {
+        if (_spell == address(0)) {
+            vm.rollFork(block.number);
+            return;
+        }
+
+        uint256 low = 5273074; // MCD_VAT Deployed Aug-06-2021 04:26:38 PM +UTC
```
iamchrissmith, line 295: I see why you picked this reference point, but why not pick a recent block.  We know that this search really only needs to look at blocks for future spells, right?

src/DssSpell.t.base.sol
```diff
@@ -266,22 +266,53 @@ contract DssSpellTestBase is Config, DssTest {
...
-            // even if mainnet has already scheduled/cast the spell
-            vm.makePersistent(address(this));
-            vm.makePersistent(address(rates));
-            vm.makePersistent(address(addr));
-            vm.makePersistent(address(deployers));
-            vm.rollFork(spellValues.deployed_spell_block);
+    }
+
+    function _rollToDeployedSpellBlock(address _spell) internal {
+        if (_spell == address(0)) {
```
gbalabasquer, line 290: Not sure I understand this part. The != address(0) check is already happening before calling this function.

src/DssSpell.t.base.sol
gbalabasquer: Yeah, taking into account this is for future spells we could put a recent block here.

src/DssSpell.t.base.sol
```diff
@@ -266,22 +266,53 @@ contract DssSpellTestBase is Config, DssTest {
...
+            vm.rollFork(block.number);
+            return;
+        }
+
+        uint256 low = 5273074; // MCD_VAT Deployed Aug-06-2021 04:26:38 PM +UTC
+        uint256 high = block.number;
+        uint256 mid;
+
+        while (low < high) {
+            mid = (low & high) + (low ^ high) / 2; // rounds down
```
gbalabasquer, line 300: Why we say that the simple `(low + high) / 2;` doesn't round down?

src/DssSpell.t.base.sol
```diff
@@ -266,22 +266,53 @@ contract DssSpellTestBase is Config, DssTest {
     function setUp() public {
         setValues(address(chief));
 
-        spellValues.deployed_spell_created = spellValues.deployed_spell != address(0) ? spellValues.deployed_spell_created : block.timestamp;
+        if (spellValues.deployed_spell != address(0)) {
+            spell = DssSpell(spellValues.deployed_spell);
+            if (spell.eta() != 0) {
```
gbalabasquer, line 271: Hmmm isn't there an issue here? We are removing the created timestamp as we assume it will be always `block.timestamp` (as we are rolling to the deployment block). However if the spell was deployed but not scheduled, this condition won't be true, then there won't be a roll to the deployment block, meaning that we will be assuming an incorrect timestamp as the deployment of the spell.
The immediate solution is removing this `if` rolling always that exists a deployed spell.

src/DssSpell.t.base.sol
naszam: yes, it rounds down to the closest integer, the other way it rounds down towards zero, truncate (I should have specified that).

gbalabasquer: Didn't understand what you mean. Rounds towards zero?

gbalabasquer: Maybe with some concrete examples for both ways I can understand what you mean.

naszam, line 300:
```suggestion
            mid = (low & high) + (low ^ high) / 2; // rounds down to zero, truncate
```

naszam:
```suggestion
            mid = (low & high) + (low ^ high) / 2; // rounds down to zero, truncate
```

gbalabasquer: ???

naszam: see [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.6.0/contracts/utils/math/Math.sol#L29)

gbalabasquer: Yeah ok, that's a complete different explanation from what you mentioned above.
It doesn't have anything to do with one "rounding down to the closest integer" vs the other "rounding down towards zero, truncate" (as they both mean the same exact same thing).
They both do the exact same thing, rounds down to the closest integer or rounds down towards zero.
The only difference is one avoids a possible overflow. So let's try to understand the reasons before doing a change and explain it.

In this specific case, the overflow won't happen in practice, so maybe it is just better to keep the simpler form. To be honest I would need to start digging in order to understand why this math does the same thing (just glancing I don't get it). Maybe as it is part of a known library we can just leave it.

naszam: testing on remix I got both implementation rounding towards zero or truncate and checking [solc docs](https://docs.soliditylang.org/en/v0.8.16/types.html#division) that's seems the case, maybe the issue I was having here was overflow.

gbalabasquer: how do you have an overflow summing two block numbers?

naszam: so, testing again against the blocks above, where I noticed the strange pattern (pass, fail, stuck) it seems to pass now with both implementation, so the issue was probably the comparison and the mid bump off by one, but anyway, I think the proposed implementation looks better for general purposes, in case the binary search pattern will be reused.

src/DssSpell.t.base.sol
iamchrissmith: This was added so that during the time after deployment before scheduling, spell crafters and reviewers do not need a local archive node. Once the spell is scheduled the local tests without an archive would fail (already scheduled), with this they will fail because forge can't `rollFork`.  But I see your point about the later timestamp tests.  Maybe we need to keep that piece in the config?

src/DssSpell.t.base.sol
iamchrissmith: I created this to address this issue: https://github.com/makerdao/spells-goerli/pull/168

src/DssSpell.t.base.sol
gbalabasquer: Right, if we want to keep this testing without archive, then we will need to do so, put back the timestamp in the config file. But doing that, I almost feel that all the binary search is less worth it and we should just keep the timestamp and block in the config.

iamchrissmith: maybe it would be worth it to see if the deploy script could automatically update with the spell address, timestamp, block number and then keep those values hardcoded?

gbalabasquer: Hmmm you mean that the script reads the spell address from the config and looks for the timestamp and block and then saves it in the same config file?

naszam: I think, following #168 pattern, the script could save the block number and timestamp into a file and then the tests will read via foundry cheatcode the values depending on the conditions, rolling back to deployed block/timestamp once scheduled/cast to allow the CI to pass.

gbalabasquer: Yeah, saving in another file and having foundry to read it is an option, requires an extra file, but might be cleaner.

iamchrissmith: i was trying to suggest that our `make deploy` script could write to the `src/test/config.sol` file after deploying/verifying and add the spell address, block, and timestamp rather than doing the search or forge script/file thing

gbalabasquer: Ok yeah, how I imagined it, might be a bit annoying to parse the file and save it in the correct place but once it's done it saves a file.
The thing is how we will get the block from the address, we need a service that provides that.

naszam: Hmm, I'm not sure about this approach as if the `config.sol` changes it might break it

gbalabasquer: Don't really follow any of both things.
1) What do you mean with breaking it?
2) why would ease the review if the config file diff would be exactly like right now excepting we don't need to fill in the change manually?

naszam: 1. depends on how you parse `config.sol` in the script, then if the `config.sol` changes in a way that could break the parsing, the script would need to be fixed, adding more overhead.
2. ease the review as the data will be in a dedicated file, like for `DssExecLib.address`, easy to inspect and modular in a way that allows to be reused for other automation applications eventually

gbalabasquer: 1. Yeah, I guess it should be smartly designed to do a good parsing that can not easily break. I guess if we look a block number and timestamp keyword shouldn't give much possibility to break.
2. I don't think I agree with this, I believe it is easier to review having all the deployed spell values together than having to go to separate files.

naszam: 1. yeah, key-value could work here but it could also break if the key will be altered somehow, so just something to keep in mind.
2. I mean having all deploying data (address, timestamp, block) in a json file or similar.
but yeah, not opposing to either of approaches, just considering pros-cons of both and see which one could be more solid and low maintenance, to keep it as simple as possible.
with @iamchrissmith approach, I guess a nice add on could be committing the change as part of the script run, like `dapp init`.

naszam: I found a simple way to parse and replace via key-value using sed, but I'm wondering how we can integrate it with the deploy script by getting those deploy info as currently we have a separate script for them that requires tx hash as input.

naszam: See #170 for a PoC

naszam: If we find an elegant way to merge those together we could automate the all process by also adding a nice add-on to create the commit for the change to just push

src/DssSpell.t.base.sol
brianmcmichael: Block is where the vat was deployed on goerli. Binary search is `O(log n)` so it doesn't matter whether we use this or block 0, the search time will not differ much.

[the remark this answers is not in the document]
brianmcmichael: I avoided max and min due to higher possibilities of collision with global variables.

src/DssSpell.t.base.sol
brianmcmichael: Reusability.

src/DssSpell.t.base.sol
brianmcmichael: I'm not sure why we need keep the time and blocks here. If you're testing locally everything works and this skips the part where it needs archive data.

naszam: yeah, the block we can rely on the binary search, on the time I'm not sure as @gbalabasquer pointed out there seems to be an issue removing it

src/DssSpell.t.base.sol
brianmcmichael: A general version of this could probably be moved into dss-test

src/DssSpell.t.base.sol
gbalabasquer: The issue is in the first message of the thread. There is a specific case that this will produce failures (spell deployed but not scheduled, which is actually a very important case in our executive process). I haven't tested it though but I'm pretty sure it will be the case.
Then we started to discuss on solutions, but those solutions are basically dropping the binary search.

gbalabasquer: An alternative solution is always requesting an archive node, which IMO is a super expensive price to pay just for avoiding to complete two values in the config.

naszam: I just tested that case and I confirm `testGeneral` Fails testing on block after the deployment one (via `make test block=8451435`)

brianmcmichael: Fast nodes do have some archive state, but only the most recent 128 blocks... this won't be suitable for our purposes even if we cache the block number. Will need some further thought, maybe the solution, instead of testing the exact expiration time, would be to modify the test to ensure that the expiration is `lt now + expiration_threshold` instead of `eq deployed time + expiration threshold`

brianmcmichael: Proposed fix here https://github.com/makerdao/spells-goerli/pull/172/files

gbalabasquer: Hmmm aren't we being less specific with that test though? I mean we are putting a weaker check.
On the other hand I was more worried about all the tests that check the specific cast time and so, aren't those failing as well?

PR #168, opened 2023-02-13
Infoless deploy test caching

iamchrissmith:
use Forge to cache the previous spells block so we start from a closer start point for our next search


src/DssSpell.t.base.sol
```diff
@@ -292,21 +292,26 @@ contract DssSpellTestBase is Config, DssTest {
             return;
         }
 
-        uint256 low = 5273074; // MCD_VAT Deployed Aug-06-2021 04:26:38 PM +UTC
+        string memory spellBlockPath = "./spellblock.txt";
+
+        uint256 low = vm.parseUint(vm.readLine(spellBlockPath));
```
naszam, line 297: wondering if the `spellblock.txt` file should be closed after reading via `vm.closeFile` cheatcode to reset the offset and allow to read it from the beginning with `vm.readLine` following [docs](https://book.getfoundry.sh/cheatcodes/fs?highlight=cheat#file-cheat-codes).

src/DssSpell.t.base.sol
```diff
@@ -292,21 +292,26 @@ contract DssSpellTestBase is Config, DssTest {
...
+                mid = (low & high) + (low ^ high) / 2; // rounds down
+                vm.rollFork(mid);
+                if (_isContractDeployed(_spell)) {
+                    high = mid;
+                } else {
+                    low = mid + 1;
+                }
             }
+            vm.writeFile(spellBlockPath, vm.toString(low));
+            vm.rollFork(low);
```
naszam, line 313: I wouldn't add more complexity here if it the change doesn't introduce practical benefits.
This seems to lock in the deployed block in the file, which is nice, but then doesn't seem to make a good use of the binary search, that will jump most of the time, if not always to the deployed block, which is fine but still doesn't convince on adding extra complexity here.
Definitely I see the proposed pattern to benefits from other applications where we might want to update dynamically a specific file based on some logic, but here I would keep it as simple as possible for now.

src/DssSpell.t.base.sol
iamchrissmith: I see your point, but since its just a one line file and we only read once here, I'm not sure we need to add this.

src/DssSpell.t.base.sol
iamchrissmith: I wasn't sure if you were talking about the overall "caching" strategy or if you were talking about checking the low block first, but I did some real rough comparisons for with/without the low block check and if statement.  

Without this code, it seems like the search takes about 10 sec longer as it searches for the block where we already found the spell (i.e. `low`).  This held true for both when there was caching and no caching (tested only the `testGENERAL` test). 

Based on this, I do think updating our block so we are closer to the one we want to find is valuable.  I understand the search is great as is and doesn't really need caching, but I think you are missing the cost of looking at the chain state  at a bunch of different points.  Essentially, the search takes an efficient number of steps, but the steps themselves are still expensive, so if we can reduce the number of them we take we are in a better spot.  plus as I mentioned in chat, searching extra blocks generates cache states in forge which just seems messy and unnecessary.

iamchrissmith: > maybe considering bumping a bit the hard-coded low/min block.

I'd rather have this automated and consistent rather than having a process where we have to do something at arbitrary points in time

src/DssSpell.t.base.sol
naszam: yeah, I mentioned for correctness, not sure what would happen when it will get read again.

src/DssSpell.t.base.sol
naszam: I was meaning to hardcode a recent block for `low/min`, not opposed to optimisations of binary search but as I said I would keep it as simple as possible.

iamchrissmith: we can see what others say, but since this is showing improved speed on my end and having fast tests is a huge benefit I think this is a better approach than just trying to remember to manually update an arbitrary block.  this piece of code is not going to change much so some added complexity in exchange for overall faster tests in my opinion is worth it

brianmcmichael: The caching is probably fine here, it's really only going to be used in CI tests, and skipping the search is probably a useful optimization there, we could revert back to a broader scan if we ever have need for it.

brianmcmichael (approved):

PR #170, opened 2023-02-20
Improve `get-deploy-info.sh`

naszam:
## Rationale

Following #166 discussion on parsing `config.sol` to replace deployed block and timestamp, the current proposed change shows how the script could be improved to automatically find and replace those values via key-value parsing using `sed`.

In this particular case `get-deploy-info.sh` have been edited to include such improvement but ideally this could be integrated directly into the `deploy.sh` script.

Opened for discussion and as PoC for the `sed` component that could be reused elsewhere as part of the spell crafting and reviewing automation process.

Tested with archived spell:
`make deploy-info tx=0x6728514cf65854f3c7efe63066b843abe443eab8930122e1e93e01b4806117e0`

Output:
```
./scripts/get-deploy-info.sh tx=0x6728514cf65854f3c7efe63066b843abe443eab8930122e1e93e01b4806117e0
Network: goerli
config.sol updated with deployed timestamp and block
```


naszam:
The logic in the PR have been moved in #171.

PR #172, opened 2023-02-24
Infoless deploy test timestamp test

brianmcmichael:
Fixes issue described at https://github.com/makerdao/spells-goerli/pull/166#pullrequestreview-1297936100 without need for timestamp.


brianmcmichael:
This loosens up the test a little bit by not requiring that we're on the exact block a spell has been deployed on, but that it will still expire within the expiration threshold. Since this is the only test that needed the exact timestamp I think it still provides sufficient assurance that the spell expiry is ok.

iamchrissmith (approved):
This is fine for me since it still proves we didn't create a long lived spell, but it does take away the protection of a short lived spell.  I don't think there is a risk there, so just noting it

naszam:
@brianmcmichael I think we can close this PR as `timestamp` will remain for now and consumed by the spell automation scripts and integration tests

PR #174, opened 2023-03-03
`rates.sh`

naszam:
## Rationale

The `rates.sh` is been introduced to ease spell reviews and to help maintaining and diff-checking IPFS rate tables.
The script is adapted from `mcd-duty` (see [mcd-cli](https://github.com/makerdao/mcd-cli/blob/master/libexec/mcd/mcd-duty)) and the pattern will be reused for the spell crafting/reviewing automation endeavour.

### Usage
The script will return the rate correspondent to the pct passed and if no pct is specified will list all the rates from 0 to 100%, with a granularity of 0.01%.

`make rates pct=0.75`
```
./scripts/rates.sh 0.75
0.75%: 1000000000236936036262880196
```

`make rates`

```
0.00%: 1000000000000000000000000000
0.01%: 1000000000003170820659990704
0.02%: 1000000000006341324285480111
0.03%: 1000000000009511510939859271
0.04%: 1000000000012681380686500226
0.05%: 1000000000015850933588756013
0.06%: 1000000000019020169709960675
0.07%: 1000000000022189089113429267
0.08%: 1000000000025357691862457863
0.09%: 1000000000028525978020323563
0.10%: 1000000000031693947650284507
0.11%: 1000000000034861600815579870
0.12%: 1000000000038028937579429884
...
```


scripts/rate.sh
```diff
@@ -0,0 +1,31 @@
...
+
+if [[ -z $1 ]];
+then
+    for n in $(seq 0 0.01 100);
+    do
+        rate "$n"
+    done
+fi
+
+if [[ $1 =~ ^([0-9]+)?([.][0-9]+)?$ ]];
```
godsflaw, line 25: Your current regex is matching an empty string.  So if you run this with no arguments, it prints the list, then tosses an error after: 
```
Parse error: bad token
    <stdin>:1
```

This suggestion should do what you want.
```suggestion
if [[ $1 =~ ^([.]|[0-9]+)([.][0-9]+)?$ ]];
```

naszam: oh! great catch! I took it from `mcd-duty` and testing I didn't noticed that edge case.

naszam: I've improved the regex in 3450967 where it should match a non zero digit followed by zero or more digits with an optional decimal part with one or more digits.

scripts/rate.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+### rate -- list all rates or compute specific ones
+### Usage:
+### ./rate.sh list all rates
+### ./rate.sh <entry> return the computed rate
```
iamchrissmith, line 6: this comment is confusing since the make command requires `pct=<entry>`

scripts/rate.sh
```diff
@@ -0,0 +1,31 @@
...
+    do
+        rate "$n"
+    done
+fi
+
+if [[ $1 =~ ^([1-9][0-9]*|[0-9])(\.[0-9]+)?$ ]];
+then
+    rate "$1"
+else
+    echo "Please specify a percentage parameter (i.e. 4.25 == 4.25%)"
```
iamchrissmith, line 29: the `make` command expects `pct=4.25` we should probably specify that here.

iamchrissmith (changes requested):

scripts/rate.sh
naszam: yeah, this refers to the script usage as it is, the `pct` part could be specified as error message on the sanity checks

Makefile
```diff
@@ -17,3 +17,4 @@ wards              :; ./scripts/wards.sh $(target)
 time               :; ./scripts/time.sh date="$(date)" stamp="$(stamp)"
 exec-hash          :; ./scripts/hash-exec-copy.sh $(url)
 fund-pause-proxy   :; ./scripts/fund-pause-proxy.sh
+rate               :; ./scripts/rate.sh $(pct)
```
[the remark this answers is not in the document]
naszam: actually, it might be more appropriate `rates` as you suggested, as the script can handle multiple rates either by listing all rates or computing specific ones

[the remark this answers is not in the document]
naszam: done in 3dc5123

scripts/rate.sh
```diff
@@ -0,0 +1,31 @@
...
+    exit 1
+fi
```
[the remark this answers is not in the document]
naszam: done in 7d08411

scripts/rate.sh
naszam: done in 4214580

naszam:
@godsflaw @iamchrissmith I should have addressed all the review comments and resolved merge conflicts as well.
It should be ready for final testings and reviews and ported to `spells-mainnet`

Makefile
```diff
@@ -16,5 +16,4 @@ diff-archive-spell :; ./scripts/diff-archive-dssspell.sh "$(if $(date),$(date),$
 wards              :; ./scripts/wards.sh $(target)
 time               :; ./scripts/time.sh date="$(date)" stamp="$(stamp)"
 exec-hash          :; ./scripts/hash-exec-copy.sh $(url)
-fund-pause-proxy   :; ./scripts/fund-pause-proxy.sh
```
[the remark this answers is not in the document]
naszam: I double checked but the merge should be correct as I get `Already up-to-date.`

iamchrissmith (approved):
changes look good to me

scripts/rates.sh
```diff
@@ -0,0 +1,27 @@
...
+    rayte=$(bc -l <<< "scale=27; e( l($normalizedamount)/(60 * 60 * 24 * 365) ) * 10^27")
+    echo "$1%: ${rayte%.*}"
+}
+
+if [[ -z $1 ]]; then
+    for n in $(seq 0 0.01 100);
+    do
+        rate "$n"
+    done
+elif [[ $1 =~ ^([1-9][0-9]*|[0-9])(\.[0-9]+)?$ ]]; then
```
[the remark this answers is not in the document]
naszam: as discussed in chat, I've tighten a bit more the regex range to only allow 0-100 range formats including two decimal digits to avoid script misuse for the specific application of `duty` rates

iamchrissmith (approved):
lgtm

PR #175, opened 2023-03-12
[CLEANUP] remove bombshell and archive script and makefile

iamchrissmith:
Clean up PR

We don't need to keep archiving `util/Bombshell.sol` in each spell if we're not using it regularly.  This removes that and moves a copy of the fund script and Makefile to the appropriate archive


brianmcmichael:
I wonder if we shouldn't just move bombshell.sol and the script out to a new repo entirely and get rid of the references here.

kmbarry1 (approved):
LGTM

kmbarry1:
> I wonder if we shouldn't just move bombshell.sol and the script out to a new repo entirely and get rid of the references here.

What would the advantage of this be? I like keeping things that were needed for spells within the spell repos.

brianmcmichael:
> What would the advantage of this be? I like keeping things that were needed for spells within the spell repos.

This isn't really a spell action, it was just an operational dependency for a spell to work. Maybe there will be a future need to fund the PP for another reason, but this keeps DAO code visible.

I've hacked up a repo we can move to MakerDAO and then we could remove the commented action from here.

https://github.com/brianmcmichael/bombshell

The-Arbiter (dismissed):
LGTM

godsflaw (dismissed):

godsflaw (approved):

brianmcmichael:
> I'm fine with having it in a separate repo (probably with some tests too wink )

Test pushed at [3717a15638929b80c0d9356ecccbac144e147f2e](https://github.com/brianmcmichael/bombshell/commit/3717a15638929b80c0d9356ecccbac144e147f2e)

PR #179, opened 2023-03-17
PE-1204: backport March 14th executive

kmbarry1:
# Description

# Contribution Checklist

- [ ] Code approved
- [ ] Tests approved
- [ ] CI Tests pass

# Checklist

- [ ] Deploy spell to Goerli `ETH_GAS_LIMIT="XXX" ETH_GAS_PRICE="YYY" make deploy`
- [ ] Cast spell on Goerli `make spell="0x-deployed-spell-address" cast-spell`
- [ ] Run `make archive-spell` or `make date="YYYY-MM-DD" archive-spell` to make an archive directory and copy `DssSpell.sol`, `DssSpell.t.sol` and `DssSpell.t.base.sol`


kmbarry1:
Interesting that the way we do chainlog tests differs slightly between mainnet and Goerli. Should I harmonize that or is the difference either intentional or just something we don't care to fix?

src/DssSpell.t.base.sol
```diff
@@ -248,17 +248,21 @@ contract DssSpellTestBase is Config, DssTest {
...
-                _scheduleWaitAndCast(address(prevSpell));
-            }
-            else {
-                // jump to nextCastTime to be a little more forgiving on the spell execution time
-                vm.warp(prevSpell.nextCastTime());
-                prevSpell.cast();
+        address[] memory prevSpells = spellValues.previous_spells;
+
+        // warp and cast previous spells so values are up-to-date to test against
+        for (uint256 i; i < prevSpells.length; i++) {
```
The-Arbiter, line 254: Referenced in main review comment.

The-Arbiter (commented):
Overall it looks good but I have some questions:

- Using dss interfaces for VatAbstract as opposed to just using VatLike; is there a reason this was done?
- `prevSpells` structure changed to array from single value; I assume this was done due to the clustering of spells at this time. I wanted to ask whether this actually did something different or was a structural change in prep for future stuff, since now we have moved part of setValues around (to the top) but setting the values of the array is commented out. I might just be missing context here. 

Another thing to note (with no effect functionally) is that a zero length array (as is the base case `new address[](0);`) will actually not trigger the `for` loop on line 254 (comment left there so you can find it easily) at all, since technically `i` is _not_ less than prevSpells.length when i is 0 by default. Again this doesn't affect anything but we need to be aware of this if we are doing any future stuff that applies when we have previous spell as empty / 0 address.

Just some things to note. Other than this it looks fine for now.

kmbarry1:
> * Using dss interfaces for VatAbstract as opposed to just using VatLike; is there a reason this was done?

Our standard pattern is to use DssInterfaces imports for these things; this is, in fact, why DssInterfaces was created. This is also how it was done in the `mainnet` spell, so it's useful to minimize the diff if anyone is comparing to the version governance already voted it.

> * `prevSpells` structure changed to array from single value; I assume this was done due to the clustering of spells at this time. I wanted to ask whether this actually did something different or was a structural change in prep for future stuff, since now we have moved part of setValues around (to the top) but setting the values of the array is commented out. I might just be missing context here.

Structural change so people who need to cast multiple spells in the future don't have to waste time on the guts of the test infra. Backporting to Goerli just to minimize diffs. The commented out line is there for extra convenience when needed.

> Another thing to note (with no effect functionally) is that a zero length array (as is the base case `new address[](0);`) will actually not trigger the `for` loop on line 254 (comment left there so you can find it easily) at all, since technically `i` is _not_ less than prevSpells.length when i is 0 by default. Again this doesn't affect anything but we need to be aware of this if we are doing any future stuff that applies when we have previous spell as empty / 0 address.

Working as intended, again want people to have to do minimal work to enable old spell testing if needed.

ryzurdo (commented):
looks good to me, ready to deploy

The-Arbiter (dismissed):
DssSpell.sol content matches mainnet
Deployed LineMom source matches between the 2 versions
Relevant content in DssSpell.t.sol matches
Relevant content in Config.sol matches

LGTM

kmbarry1:
Deployed and archived (per discussion this last week that those are being combined to save back-and-forths), ready for final review and testing.

The-Arbiter (approved):
```rust
./scripts/diff-archive-dssspell.sh "2023-03-17"
Spell, tests and base match the archive directory 2023-03-17-DssSpell
```

Archive OK

ryzurdo (approved):
Archive Diff checks out.
Deployed spell matches GH.
Ready to cast and merge

kmbarry1:
Spell is cast.

Issue #186, opened 2023-04-19
Tests running 2x for each push


amusingaxl:
Right now we have the following config on Github test workflow:

```yaml
on: [push, pull_request]
```

This causes tests to run 2x for each push on an open PR:

![image](https://user-images.githubusercontent.com/112016538/233079012-d1888408-f7e1-48b1-81e4-84388471f57d.png)

This not only significantly slows down the finalization of the checks, but also consumes 2x the number of requests from the provider.

amusingaxl:
This is not how Github Actions work I'm afraid.

![image](https://user-images.githubusercontent.com/112016538/233361226-6f253c3e-710b-4f11-851b-8f79c935b44b.png)

Whenever there is an outstanding PR, every new commit on a branch which is the PR target will  trigger a new action, not only when the PR is created.

If you have both `push` and `pull_request` triggers, it will trigger actions 2x.

> The image above references two separate PR's, which need to be tested independently of each other.

It does not, it's the same PR. Look closely.

amusingaxl:
Closing this issue for now.

PR #187, opened 2023-04-19
Optimize Github Actions test workflow

amusingaxl:
This is a fix for #186.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,6 +1,12 @@
 name: Tests
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
 
 jobs:
   tests:


brianmcmichael:
CI works as intended.

amusingaxl:
I'm afraid it's not doing what you think it should.

gbalabasquer:
The `push` checks what the commit does to the branch where we are working on. The `pull_request` checks what the commit does to the PR once if would be merged to the base at the moment.
In general they will run the same, but it is not always the case. If for any reason two PRs are worked in parallel, the second to be merged is interested to check both. The `push` because it is the actual code that will be deployed as the spell and the `pull_request` because it will be how it will be left after merging.
If this PR is merged, a vulnerability would be introduced as it wouldn't be actually checking the branch the crafter is working on (an again, which is the most trustable source for your deployed spell).
I made this small POC to show they run different things: https://github.com/gbalabasquer/test-ci/pull/3 (if you can't access to the CI details I can share screenshots, or you can test in your own repo forking both branches and making the PR).
Also if there is a conflict, the `pull_request` one will not run at all, so that already gives an insight.
As a last thing, I'll say that I'm pretty sure with this PR, there won't be any CI at all running when someone makes a PR from a non master branch to another non master branch (haven't test it that though).

amusingaxl:
I don't think I get exactly the problem you are trying to avoid here.

From Github Docs, we have:

> Runs your workflow when activity on a pull request in the workflow's repository occurs. For example, if no activity types are specified, the workflow runs when a pull request is opened or reopened or **when the head branch of the pull request is updated**.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

A commit made to pull request branch will **always** update its head, meaning the workflow will run for every push already.

> The pull_request checks what the commit does to the PR if would be merged to the base at the moment.

From the `pull_request_target` docs we have:

> This [`pull_request_target`] event runs in the context of the base of the pull request, **rather than in the context of the merge commit**, as the pull_request event does.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

I believe I got what you are trying to say.

> In general they will run the same, but it is not always the case. If for any reason two PRs are worked in parallel, the second to be merged is interested to check both. The push because it is the actual code that will be deployed as the spell and the pull_request because it will be how it will be left after merging.

Under which circumstances would this happen? Parallel spells? What would be the applicability here?

In any case, isn't this something you want to avoid?

Even if the spell tests run in parallel, in the end the tests run in the same environment and there should be an order of execution. This effectively invalidates the test run on `push` on the 2nd spell, because the system is not taking into account the 1st spell.

If one of the PRs is merged first, the target (i.e.: `master`) branch should be merged back into the 2nd PR (this can be enforced IIRC), which would trigger the `pull_request` action one more time.

I know about the [`previous_spells`](https://github.com/makerdao/spells-goerli/blob/6ef52322370bd219368e7941db0080f2181a65e6/src/test/config.sol#L96) mechanism to ensure ordering.

So if that's the case and the ordering is guaranteed by `previous_spells`, making the 2nd spell to have to account for any changes introduced by the 1st spell, I can't see the value of having the `push` trigger for the 2nd one.

If the crafter (purposely or unwillingly) does not make use of the `previous_spells` mechanism, we could have the `push` trigger pass the tests, but in reality the execution would fail. or the other way around.

> Also if there is a conflict, the pull_request one will not run at all, so that already gives an insight.

Yes, but we can enforce this check to be run, which would effectively prevent conflicted PRs from being merged.

> As a last thing, I'll say that I'm pretty sure with this PR, there won't be any CI at all running when someone makes a PR from a non master branch to another non master branch (haven't test it that though).

You are correct in your assumption. This is deliberate because the `master` branch is supposed to be the stable one. I get that the workflow for spells are a little bit different, since spells are deployed from their branch and not from `master` though. The `pull_request` trigger could not have a restriction on branches, so it would work for any PR.

However, the main point I'm questioning here is the need for a `push` trigger.

Please let me know if I missed something.

I'm not trying to be a pain here, but since we are inheriting this process, we need to understand it in as many details as possible.

brianmcmichael:
This spell is not the time to fix things or aim for innovation or cleverness. Stick to existing patterns until you learn how the existing process works.

amusingaxl:
I am not trying to be clever neither I implied this PR should be merged DURING this spell process, it's simply here as a placeholder for further discussion with the people that will be involved in this process, since this used to be a HIGHLY gatekeeped process.

I am just asking questions, without making any assumptions, so you can stop this condescending dismissive comments because they are surely not beneficial to anyone involved, including yourself.

brianmcmichael:
> This spell is not the time to fix things or aim for innovation or cleverness. Stick to existing patterns until you learn how the existing process works.

Sorry, this comment was on the wrong PR, I'll let Gonza reply on this PR.

gbalabasquer:
Sorry, I forgot to reply about this:

> Under which circumstances would this happen? Parallel spells? What would be the applicability here?

More probably a feature update PR (for tests for example) running at the same time than an Executive PR.

> Even if the spell tests run in parallel, in the end the tests run in the same environment and there should be an order of execution. This effectively invalidates the test run on `push` on the 2nd spell, because the system is not taking into account the 1st spell.

Two spells is probably a case that won't exist, or maybe yes as it was happening this week. But it is true that after merging to `master` the first spell we expect the crafter will merge `master` in the second PR.

> You are correct in your assumption. This is deliberate because the `master` branch is supposed to be the stable one. I get that the workflow for spells are a little bit different, since spells are deployed from their branch and not from `master` though. The `pull_request` trigger could not have a restriction on branches, so it would work for any PR.
> 
> However, the main point I'm questioning here is the need for a `push` trigger.

Well that is exactly my point, in this case `push` inside branches it is super important as it is the one that you are actually basing to build, deploy and test the spell. Running the CI based on how your PR would run once being merged to `master` might hide things that you wanted actually to check.
In this case, the `pull_request` CI run will be less important than the branch `push` one, it will be just to make sure that when it is merged to `master `the CI won't break. However being less important it doesn't mean that we shouldn't remove it IMO.

As I previously mentioned, this PR also discards the non master to non master PRs, basically disable any CI run completely. Why would we like to do so? Why do we think it isn't important to know if tests still pass when for example someone makes a PR for a Spell Executive PR?

amusingaxl:
> Well that is exactly my point, in this case push inside branches it is super important as it is the one that you are actually basing to build [...]

My point is that this should NOT happen... there shouldn't be a build from a stale branch, because you can't ensure the actual code being executed in the spell will work in case there are conflicting changes.

> Running the CI based on how your PR would run once being merged to master might hide things that you wanted actually to check.

Any specific scenarios come to mind? 

I believe that could be the case for an environment where things run in parallel, however spells always run sequentially. Not to repeat myself, but the test executed on `push` for a spell which is behind `master` after another spell has no validity, because the environment WILL change for sure.

> As I previously mentioned, this PR also discards the non master to non master PRs, basically disable any CI run completely.

Yes, this part is true, that's why I moved it back to draft, but this can basically be changed to:

```yaml
on:
  push:
    branches:
      - master
  pull_request:
```

That would trigger it for any open PR on any branch.

However, I see that by the time a spell is merged into `master`, the spell has already been cast, which means the CI test will always fail.

If commits directly into `master` are disabled, we could simply not run tests on push for it. I haven't tested it, but I believe this would do the trick:

```yaml
on:
  push:
    branches:
      - !master
  pull_request:
```

---

To give you some context, we're trying to negotiate a plan with Alchemy to shorten the feedback loop for spell tests. Any improvements in this process that doesn't impact security is welcome to be mindful of the costs.

brianmcmichael:
You can shorten the feedback loop by setting an `ETH_RPC_URL` in your local environment and using `make test` before you push code to a public repo. You should not be using the CI as a development environment. This PR should be closed.

amusingaxl:
You are completely missing the point here.
Of course, I run tests before pushing, I am not an idiot.

I'm thinking about the reviewer feedback loop, not the crafter.

It's getting increasingly harder to run nodes locally, even for Goerli. I can do that because I own a beefy machine, however it's not feasible to require that everyone involved in spells to run their own nodes, specially now that there will be more people involved.

Local nodes come with several problems, local dependencies can become outdated, there's also the spell crafting/reviewing learning curve itself, so a fast and reliable CI is the minimum we should provide to the people to make this productive.

amusingaxl:
Closing this for now.

PR #194, opened 2023-05-02
Fix: `diff-deployed-dssspell` is breaking

### LINKED ISSUE #210: `diff-deployed-spell` script requires some improvements
I noticed that sometimes the Etherscan API returns the source code with Windows-style line-breaks (`\n\r`).
This would break the diff in any Unix-based system.

Also, bash has some slightly different implementations across MacOS and Linux, so it's a good idea to avoid the shell trying to be funny with strings.

---

Another problem is that sometimes the only difference between the repo and the spell code is a trailing new-line.

That would cause `diff` to exit with a non-zero code, making the script fail, which would prevent this script to be used in tandem with any other scripts following it, even if the diff is fine.

Cross-posted from [spells-mainnet#352](https://github.com/makerdao/spells-mainnet/issues/352).

amusingaxl:
I noticed that sometimes the Etherscan API returns the source code with
Windows-style line-breaks (`\n\r`). This would break `diff` in any
Unix-based system.

Also, `bash` has some slightly different implementations across MacOS and
Linux, so it's a good idea to try to avoid the shell trying to be funny
with strings. For that reason, I changed the Etherscan source download
step to store the output in a file instead of a string in memory.

Since everything is erased at the end with `make clean`, I believe that
wouldn't cause an issue.

The final issue this commit fixes is that sometimes the only difference
between the repo and the spell code is a trailing new-line. That would
cause `diff` to exit with a non-zero code, making the script fail.

This would prevent this script to be used in tandem with any other
scripts following it, even if the diff is fine.

[Fixes #210]


The-Arbiter:
Agreed that the trailing newline is an issue (it was for PE MacOS users). This script is relatively new so I will take a look at how this functions and whether there are any considerations we have. I'm still manually diffing deployed spells rather than migrating to this script so this will be good, I'll look in more detail soon.

amusingaxl:
> Edit: If there is a way to remove jq as well I am interested in that as it is not a default install on some platforms (like MacOS)

I'd definitely NOT try and parse JSON manually. However, it may be a good thing to check if the users have `jq` installed, and point them in the right direction in case they do not, than simply assuming it's installed.

scripts/diff-deployed-dssspell.sh
```diff
@@ -2,23 +2,31 @@
...
 
+command -v jq &> /dev/null || { echo '`jq` command could not be found. Please install it: https://github.com/stedolan/jq'; exit 1; }
+[[ "$ETHERSCAN_API_KEY" ]] || { echo -e "Please set ETHERSCAN_API_KEY"; exit 1; }
+
 if [[ "$1" =~  ^0x[0-9a-fA-F]{40}$ ]]; then
     deployed_spell_address=$1
 else
     # Read contract address from config.sol
-    deployed_spell_address=$(grep -oE 'deployed_spell:\s+address\((0x[a-fA-F0-9]+)\)' "src/test/config.sol" | sed -E 's/^.*\((.*)\)/\1/')
+    deployed_spell_address=$(grep -oE 'deployed_spell:\s*(address\((0x[a-fA-F0-9]{40})\)|(0x[a-fA-F0-9]{40}))' "src/test/config.sol" | sed -E 's/.*(0x[a-fA-F0-9]{40}).*/\1/')
```
SidestreamColdMelon, line 12: Nice catch on the new `|` operator in `address\((0x[a-fA-F0-9]{40})\)|(0x[a-fA-F0-9]{40})`, although there was just one spell like this 🙂

SidestreamColdMelon (approved):

0xdecr1pto (approved):

PR #203, opened 2023-05-31
Add spark spell contents and tests

hexonaut:


0xdecr1pto:
@hexonaut As per exec sheet, we need to add Spark Proxy to PSM. Do you prefer to do that in this PR?

hexonaut:
The ESM? It's been added with the authorize line.

src/DssSpell.sol
```diff
@@ -136,8 +148,15 @@ contract DssSpellAction is DssAction {
...
 
         // --- Spark Protocol Parameter Changes ---
-        // Poll: https://vote.makerdao.com/polling/QmWatYqy#poll-detail
+        // D3M Parameter Adjustments Poll: https://vote.makerdao.com/polling/QmWatYqy#poll-detail
+        // Executive Proxy Poll: https://vote.makerdao.com/polling/Qmc9fd3j#poll-detail
+        // Onboard rETH Poll: https://vote.makerdao.com/polling/QmeEV7ph#vote-breakdown (Inside Proxy Spell)
+        // DAI Interest Rate Strategy Poll: https://vote.makerdao.com/polling/QmWodV1J#poll-detail (Inside Proxy Spell)
         // Forum: https://forum.makerdao.com/t/2023-05-24-spark-protocol-updates/20958
+        DssExecLib.setIlkAutoLineParameters("DIRECT-SPARK-DAI", /* line */ 20 * MILLION, /* gap */ 20 * MILLION, /* ttl */ 8 hours);
+        DssExecLib.authorize(SPARK_PROXY, DssExecLib.esm());
```
SidestreamColdMelon, line 157: if we're authorizing `SPARK_PROXY` on ESM, shouldn't it be added to chainlog? I would then suggest to name it as `SUBDAO_PROXY_SPARK` to not confuse with other proxies in the chainlog and be able to filter them in the future

SidestreamSweatyPumpkin:
Looks good, just one question (marked with ❓ in the checklist below). Will explicitly list it below this paragraph as well. Checklist below is the goerli-reviewer's checklist's subset (the parts that i find relevant).

- ❓ Should we add spark proxy to the chainlog as per checklist? (same q as in the comment above though :D) 


Checklist 
---

* [ ] Static Interfaces
  * [ ] ensure they match `dss-interfaces`
    external interfaces are used, this checkbox does not seem relevant
  * [x] check on-chain interface of deployed contract via `cast interfaces <contract_address>` to ensure correctness
  * [x] interface naming style should match with `Like` suffix (e.g. `VatLike`)
  * [x] ensure they only list used functions in spell code
* [ ] External Contracts Calls (e.g. Starknet)
  * [x]  Target Contract don't block spell execution
  * [ ]  External call is NOT delegate call
    - it is, skipping this point as per spell requriements
  * [x]  Target Contract doesn't have permissions on the Vat
    - proxy does not have this https://goerli.etherscan.io/address/0x3068FA0B6Fc6A5c998988a271501fF7A6892c6Ff#code
    - spell does not have  this 0x4e847915D8a9f2Ab0cDf2FC2FD0A30428F25665d
    - acl manager does not have this 0xb137E7d16564c81ae2b0C8ee6B55De81dd46ECe5
  * [x]  Target Contract doesn't do anything untoward (e.g. interacting with unsafe contracts)
    - reviewing the actions of the contracts resulted into me not noticing malicious things
  * [x]  MCD Pause Proxy doesn't give any approvals
    - reviewing the actions of the contracts resulted into me not noticing this
  * [x]  Spark proxy can only be called via Pause proxy 
     - deployer denied 
     - PP relid
  * [ ]  All possible actions of the Target Contract are documented
     - cannot ensure this since i dont have access to full documentation (i can only see some comments on etherscan but dont have access to web version of the docs.)
       seems fine though
  * [ ]  Target contract is not upgradable
    there's a proxy that can execute anything, i would not check this box.
  * [ ]  Target Contract is included in the ChainLog
     ❓should we do this with spark proxy
  * [x]  Test Coverage is comprehensive
* [x] Ensure every spell variable is declared as public/internal
* [x] DssExeclib used
   - [x] risk parameter changes
   - [x] authorization
* [x] Tests PASS
  * [x] Ensure Good Coverage
  * [x] Ensure every test function is declared as public if enabled or private if disabled

src/DssSpell.sol
hexonaut: Good point - yeah this makes sense.

hexonaut:
added chainlog

SidestreamSweatyPumpkin (approved):

SidestreamSweatyPumpkin (approved):

PR #205, opened 2023-06-05
Fix: `make test-forge match="<test>"` is broken

amusingaxl:
The [`--match` flag was finally removed](https://book.getfoundry.sh/reference/forge/forge-test#test-options) from `forge` in favor of `--match-test`/`--match-contract`/`--match-path`.

`make test-forge` script is now broken.


The-Arbiter (approved):
LGTM, simple change

0xdecr1pto (approved):

PR #209, opened 2023-06-19
[Mainnet Sync] Fix misleading comment on `testNewChainlogValues`

### LINKED ISSUE #211: Missing explicit check for chainlog version on base tests
In Goerli, we currently [do not check for the version in `src/test/config.sol`](https://github.com/makerdao/spells-goerli/blob/bd3c2de0b0527a79691aa2b2ba22a04536a42c28/src/DssSpell.t.base.sol#L1767-L1776) the same way [we do for Mainnet](https://github.com/makerdao/spells-mainnet/blob/b8965c4d0b0eafc9f92d012a08b94c0fceb640e1/src/DssSpell.t.base.sol#L1864).

That requires the `testNewChainlogValues` test to be [carried on for every spell](https://github.com/makerdao/spells-goerli/blob/46bdbdd3e3793c3cb41d064c8ce477855093d33e/archive/2023-05-10-DssSpell/DssSpell.t.sol#L233-L239) we write on Goerli, even if there are no new entries to the chainlog, only to check the chainlog version.

amusingaxl:
For more details, see [spells-mainnet#348](https://github.com/makerdao/spells-mainnet/pull/348).

[Fixed #211]


amusingaxl:
The origin of the problem is that the Goerli test suite was out of sync with Mainnet.

On Mainnet there is a `chainlog_version` on the `SystemValues` struct which is [explicitly checked](https://github.com/makerdao/spells-mainnet/blob/b8965c4d0b0eafc9f92d012a08b94c0fceb640e1/src/DssSpell.t.base.sol#L1864).

For Goerli, there wasn't such config.

I traced back the "don't disable" comment, and it dates back when PE decided to [check whether the chainlog version should be bumped or not](https://github.com/makerdao/spells-goerli/blob/46bdbdd3e3793c3cb41d064c8ce477855093d33e/archive/2023-01-24-DssSpell/DssSpell.t.base.sol#L1697-L1728).

That eliminated the explicit check for the current chainlog version.

Somehow, this change made its way to Mainnet.

SidestreamColdMelon (approved):
Thanks for tracing it down in https://github.com/makerdao/spells-mainnet/pull/348#issue-1764179144

0xdecr1pto (approved):

0xdecr1pto (approved):

SidestreamColdMelon (approved):

PR #212, opened 2023-06-20
Fix: make `AutoLine` tests enforce config cleanup

### LINKED ISSUE #189: Cleanup `RWA007-A` auto-line config params
The auto-line for `RWA007-A` [has been disabled](https://github.com/makerdao/spells-goerli/blob/6ef52322370bd219368e7941db0080f2181a65e6/archive/2023-03-28-DssSpell/DssSpell.sol#L123), however the old params in the config file [were left there](https://github.com/makerdao/spells-goerli/blob/6ef52322370bd219368e7941db0080f2181a65e6/src/test/config.sol#L1376-L1378):

image

The issue is that this does not reflect the current state of the `MCD_IAM_AUTO_LINE` contract:

image

The test didn't catch this because they [simply ignore the values if `aL_enabled` is `false`](https://github.com/makerdao/spells-goerli/blob/6ef52322370bd219368e7941db0080f2181a65e6/src/DssSpell.t.base.sol#L486-L489).

We probably need to fix the test to make it enforce the auto-line parameters are cleaned up when an ilk auto-line is removed.

amusingaxl:
Currently, the test does not enforce `AutoLine` config cleanup if `aL_enabled` is set to `false`.

This makes the config file to be out-of-sync with the on-chain data, which we want to avoid.

[Fixes #189]


0xdecr1pto (approved):

SidestreamColdMelon (approved):
No previous tests were dropped, 3 new tests were added to validate that `aL_*` parameters are set to 0 in the config if AutoLine is disabled. That qualifies this PR as an improvement 🙂

PR #217, opened 2023-07-05
Flapper test minor changes

sunbreak1211:


oldchili (approved):
LGTM

src/DssSpell.t.sol
```diff
@@ -521,7 +521,6 @@ contract DssSpellTest is DssSpellTestBase {
         stdstore.target(address(vow)).sig("Sin()").checked_write(3_000_000 * RAD);
         stdstore.target(address(vow)).sig("Ash()").checked_write(2_000_000 * RAD);
 
-        assertEq(flap.dai(), address(dai));
```
SidestreamColdMelon: What do you mean by [`No need to check dai again`](https://github.com/makerdao/spells-goerli/pull/217/commits/2108e93fbb7194e8bec4b868dd250633a3e5c976)? Where else is it checked?

sunbreak1211: It is enforced in the init process the check of `flapper.daiJoin == dss.daiJoin` (https://github.com/makerdao/spells-goerli/blob/2023-07-12-flapper-test-minor-changes/src/dependencies/dss-flappers/FlapperInit.sol#L72) and `dai` comes from the `daiJoin`. In the case of `gem` there is not validation at all in the init that the MKR token is being used.

PR #220, opened 2023-07-26
Feature: add `no-match` option to `make test-forge` / Refactor: use `make test` instead of `make test-forge` in CI

amusingaxl:
Sometimes a spell crafter wants to re-run a couple of tests only for faster execution when using an external JSON RPC provider. For that we have the `match='<regex>'` option.

However there is a case where the crafter might want to run the entire test suite, but skip only the ones that take longer to execute. This can be useful when debugging spells, shortening the feedback loop.

This PR adds the `no-match='<regex>'` options which uses the `--no-match-test` option from `forge` to skip a couple of tests.

For instance, one could skip `testAuth` and `testAuthInSources` while iterating over the spell code and only run those heavy tests at the end:

```bash
make test-forge no-match='testAuth|testAuthInSources'
```

### DIFF
--- a/scripts/test-dssspell-forge.sh
+++ b/scripts/test-dssspell-forge.sh
@@ -9,6 +9,7 @@ do
     VALUE=$(echo "$ARGUMENT" | cut -f2 -d=)
 
     case "$KEY" in
+            no-match)   NO_MATCH="$VALUE" ;;
             match)      MATCH="$VALUE" ;;
             block)      BLOCK="$VALUE" ;;
             *)
@@ -22,12 +23,16 @@ export FOUNDRY_OPTIMIZER=false
 export FOUNDRY_OPTIMIZER_RUNS=200
 export FOUNDRY_ROOT_CHAINID=5
 
-if [[ -z "$MATCH" && -z "$BLOCK" ]]; then
-    forge test --fork-url "$ETH_RPC_URL"
-elif [[ -z "$BLOCK" ]]; then
-    forge test --fork-url "$ETH_RPC_URL" --match-test "$MATCH" -vvv
-elif [[ -z "$MATCH" ]]; then
-    forge test --fork-url "$ETH_RPC_URL" --fork-block-number "$BLOCK"
-else
-    forge test --fork-url "$ETH_RPC_URL" --match-test "$MATCH" --fork-block-number "$BLOCK" -vvv
+EXTRA_ARGS=''
+
+if [[ -n "$MATCH" ]]; then
+    EXTRA_ARGS="${EXTRA_ARGS} -vvv --match-test ${MATCH}"
+elif [[ -n "$NO_MATCH" ]]; then
+    EXTRA_ARGS="${EXTRA_ARGS} -vvv --no-match-test ${NO_MATCH}"
 fi
+
+if [[ -n "$BLOCK" ]]; then
+    EXTRA_ARGS="${EXTRA_ARGS} --fork-block-number ${BLOCK}"
+fi
+
+forge test --fork-url "$ETH_RPC_URL" $EXTRA_ARGS
--- a/scripts/test-dssspell-forge.sh
+++ b/scripts/test-dssspell-forge.sh
@@ -9,8 +9,8 @@ do
     VALUE=$(echo "$ARGUMENT" | cut -f2 -d=)
 
     case "$KEY" in
-            no-match)   NO_MATCH="$VALUE" ;;
             match)      MATCH="$VALUE" ;;
+            no-match)   NO_MATCH="$VALUE" ;;
             block)      BLOCK="$VALUE" ;;
             *)
     esac
--- a/scripts/test-dssspell-forge.sh
+++ b/scripts/test-dssspell-forge.sh
@@ -23,16 +23,16 @@ export FOUNDRY_OPTIMIZER=false
 export FOUNDRY_OPTIMIZER_RUNS=200
 export FOUNDRY_ROOT_CHAINID=5
 
-EXTRA_ARGS=''
+TEST_ARGS=''
 
 if [[ -n "$MATCH" ]]; then
-    EXTRA_ARGS="${EXTRA_ARGS} -vvv --match-test ${MATCH}"
+    TEST_ARGS="${TEST_ARGS} -vvv --match-test ${MATCH}"
 elif [[ -n "$NO_MATCH" ]]; then
-    EXTRA_ARGS="${EXTRA_ARGS} -vvv --no-match-test ${NO_MATCH}"
+    TEST_ARGS="${TEST_ARGS} -vvv --no-match-test ${NO_MATCH}"
 fi
 
 if [[ -n "$BLOCK" ]]; then
-    EXTRA_ARGS="${EXTRA_ARGS} --fork-block-number ${BLOCK}"
+    TEST_ARGS="${TEST_ARGS} --fork-block-number ${BLOCK}"
 fi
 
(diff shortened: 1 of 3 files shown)


SidestreamColdMelon:
Should we maybe instead of adding `no-match` and memorizing `testAuth|testAuthInSources` add new `make test quick` option or `make test-quick` command that will hardcode `testAuth|testAuthInSources` in the makefile?

scripts/test-dssspell-forge.sh
```diff
@@ -22,12 +23,16 @@ export FOUNDRY_OPTIMIZER=false
...
+    EXTRA_ARGS="${EXTRA_ARGS} -vvv --match-test ${MATCH}"
+elif [[ -n "$NO_MATCH" ]]; then
+    EXTRA_ARGS="${EXTRA_ARGS} -vvv --no-match-test ${NO_MATCH}"
+fi
+
+if [[ -n "$BLOCK" ]]; then
+    EXTRA_ARGS="${EXTRA_ARGS} --fork-block-number ${BLOCK}"
 fi
+
+forge test --fork-url "$ETH_RPC_URL" $EXTRA_ARGS
```
The-Arbiter, line 38: Argument passthrough via an `EXTRA_ARGS` env var makes sense in terms of usability.

Security considerations here are that `make test` can now produce unexpected results since `EXTRA_ARGS` could be anything and isn't strictly limited to 'MATCH' and 'BLOCK' like before.

Consider the case where a malicious spell team member wants to make their tests pass / fail / produce certain results. A custom 'EXTRA_ARGS' setup may allow for this while still producing test results which appear genuine.

Of course, someone could always run the tests directly via 'forge test' and do this.

I propose one of two fixes here:

1) EXTRA_ARGS is echoed / provided as output which must be provided alongside passing or failing test results if only to provide context for future spell team members looking back at test results when debugging issues in the future.

2) 'EXTRA_ARGS' is limited to certain env vars which can be passed in (MATCH, NO_MATCH and BLOCK for example) which is the existing method of design.

Overall this is a net improvement in terms of utility. The only thing here is that the 'tests PASS' checklist requirement becomes open ended when the entire set of execution setting arguments are permitted. Limiting the permissible env vars which can be set (old design) is one way of solving this, another way is by changing the checklist such that it mandates `forge test --fork-url "$ETH_RPC_URL"` as the test command being used.

Open to feedback here

amusingaxl:
> Should we maybe instead of adding `no-match` and memorizing `testAuth|testAuthInSources` add new `make test quick` option or `make test-quick` command that will hardcode `testAuth|testAuthInSources` in the makefile?

Those are just examples, Dewiz is working on a blog post that digs deep into this issue, but the long-story short is that some tests are way more expensive than others.
Since spells tests are E2E, sometimes you are having issues with a specific part, so you don't want to execute all of them while debugging. It's better to leave this up to the dev.

scripts/test-dssspell-forge.sh
amusingaxl: > 'EXTRA_ARGS' is limited to certain env vars which can be passed in (MATCH, NO_MATCH and BLOCK for example) which is the existing method of design.

It's already like that... Let me add some contextual comments to the code here:

```bash
EXTRA_ARGS='' # whatever value in the EXTRA_ARGS env var is discarded

# if there's a match, add it to extra args
if [[ -n "$MATCH" ]]; then
    # extra args will contain --match-test
    EXTRA_ARGS="${EXTRA_ARGS} -vvv --match-test ${MATCH}"
elif [[ -n "$NO_MATCH" ]]; then
    # extra args will contain --no-match-test
    EXTRA_ARGS="${EXTRA_ARGS} -vvv --no-match-test ${NO_MATCH}"
fi

# if there's a block, add it to extra args
if [[ -n "$BLOCK" ]]; then
    EXTRA_ARGS="${EXTRA_ARGS} --fork-block-number ${BLOCK}"
fi
```

Notice how no other option can be added, unless someone tries to 'inject' stuff like `match='"" --other-option FOO'`, but this can happen anyway, and it's hard to prevent in pure bash script.

SidestreamColdMelon (approved):

scripts/test-dssspell-forge.sh
The-Arbiter: I agree with these points. Confirmed that the code as present here does in fact limit the env being used by the script, so there is no additional security risk IMO. This was my mistake when interpreting the way it was handling `EXTRA_ARGS`.

That being said, the naming is a bit confusing, can we please have the `EXTRA_ARGS` from env var and the `EXTRA_ARGS` used by the script using different names?

For example, `EXECUTION_ARGS` or `TEST_ARGS` or `INPUT_TEST_ARGS` would be more clear, as this is part of why I was confused initially.

scripts/test-dssspell-forge.sh
```diff
@@ -22,12 +23,16 @@ export FOUNDRY_OPTIMIZER=false
...
 
-if [[ -z "$MATCH" && -z "$BLOCK" ]]; then
-    forge test --fork-url "$ETH_RPC_URL"
-elif [[ -z "$BLOCK" ]]; then
-    forge test --fork-url "$ETH_RPC_URL" --match-test "$MATCH" -vvv
-elif [[ -z "$MATCH" ]]; then
-    forge test --fork-url "$ETH_RPC_URL" --fork-block-number "$BLOCK"
-else
-    forge test --fork-url "$ETH_RPC_URL" --match-test "$MATCH" --fork-block-number "$BLOCK" -vvv
+EXTRA_ARGS=''
```
The-Arbiter, line 26: `FILTERED_ENV_ARGS`, `PERMITTED_ENV_ARGS`, `APPROVED_ENV_ARGS` etc

scripts/test-dssspell-forge.sh
```diff
@@ -22,12 +23,16 @@ export FOUNDRY_OPTIMIZER=false
...
 
-if [[ -z "$MATCH" && -z "$BLOCK" ]]; then
-    forge test --fork-url "$ETH_RPC_URL"
-elif [[ -z "$BLOCK" ]]; then
-    forge test --fork-url "$ETH_RPC_URL" --match-test "$MATCH" -vvv
-elif [[ -z "$MATCH" ]]; then
-    forge test --fork-url "$ETH_RPC_URL" --fork-block-number "$BLOCK"
-else
-    forge test --fork-url "$ETH_RPC_URL" --match-test "$MATCH" --fork-block-number "$BLOCK" -vvv
+TEST_ARGS=''
```
SidestreamColdMelon, line 26: Do you want to use the same `local` keyword as in the [mainnet PR](https://github.com/makerdao/spells-mainnet/pull/358)?
```suggestion
local TEST_ARGS=''
```

scripts/test-dssspell-forge.sh
amusingaxl: I went with `TEST_ARGS`... it feels weird calling this `ENV_` because they are not environment variables, they are command line arguments.

scripts/test-dssspell-forge.sh
amusingaxl: Already reverted it.

SidestreamColdMelon (approved):

SidestreamColdMelon (dismissed):

The-Arbiter (approved):
Changes LGTM, issues I raised previously appear to have been addressed.

Maintenance PR OK to merge

The-Arbiter (approved):
Changes OK.

Approved alongside corresponding mainnet repo PR https://github.com/makerdao/spells-mainnet/pull/358/files

LGTM

SidestreamSweatyPumpkin (approved):

PR #225, opened 2023-08-16
Improvement: script to cast spell on tenderly

SidestreamColdMelon:
### Description

This PR adds `node.js`-based script to create [tenderly forks](https://docs.tenderly.co/simulations-and-forks/forks) and execute pre-deployed spell on them. Currently it's optional to use, but later on we might want to include it into the checklist. The script is inspired by [a similar functionality](https://github.com/marsfoundation/spark-spells/blob/master/tenderly.js) used by the Spark team

### Usage instructions

1. Create Tenderly account (no trial period needed atm) https://dashboard.tenderly.co/register
    - Note down `TENDERLY_USER` and `TENDERLY_PROJECT` values
2. Create Tenderly access token https://dashboard.tenderly.co/account/authorization
    - Note down `TENDERLY_ACCESS_KEY` values
3. Export env vars specified above or create `scripts/cast-on-tenderly/.env` file with them
4. Execute `make cast-on-tenderly spell=0x...`, with the address of the actual spell that hasn't been casted yet
    - Eg `make cast-on-tenderly spell=0x341281316C53a6c9b099581C9f87665FA5815090` for the current spell
    - The execution should finish with `successfully casted`
6. Open the `publicly sharable url` printed into the console (it should require no credentials)


scripts/cast-on-tenderly/index.js
```diff
@@ -0,0 +1,87 @@
+import 'dotenv/config';
+import axios from 'axios';
+import { Contract, ethers } from 'ethers';
+
+const NETWORK_ID = 5;
+const CHIEF_ADDRESS = '0x33Ed584fc655b08b2bca45E1C5b5f07c98053bC1';
```
amusingaxl, line 6: Wouldn't it be better to have the `Chainlog` address hard-coded here and fetch `MCD_CHIEF` from it?

Makefile
```diff
@@ -13,6 +13,7 @@ flatten              :; hevm flatten --source-file "src/DssSpell.sol" > out/flat
 diff-deployed-spell  :; ./scripts/diff-deployed-dssspell.sh $(spell)
 check-deployed-spell :; ./scripts/check-deployed-dssspell.sh
 cast-spell           :; ./scripts/cast-dssspell.sh $(spell)
+cast-on-tenderly     :; cd ./scripts/cast-on-tenderly/ && npm i && node index.js $(spell)
```
amusingaxl, line 16: Maybe switch back to the original directory after executing:

```suggestion
cast-on-tenderly     :; cd ./scripts/cast-on-tenderly/ && npm i && node index.js $(spell); cd -
```

scripts/cast-on-tenderly/index.js
```diff
@@ -0,0 +1,87 @@
...
+
+    const provider = new ethers.providers.JsonRpcProvider(rpcUrl);
+    const signer = provider.getSigner();
+
+    console.info('getting the hat...');
+    await provider.send('tenderly_setStorageAt', [
+        CHIEF_ADDRESS,
+        ethers.utils.hexZeroPad(ethers.utils.hexValue(CHIEF_HAT_SLOT), 32),
+        ethers.utils.hexZeroPad(SPELL_ADDRESS, 32),
+    ]);
```
amusingaxl, line 56: Could we try to impersonate the pause proxy instead of relying on `setStorageAt`?
Seems doable at first sight: https://docs.tenderly.co/simulations-and-forks/reference/tenderly-fork-customization-via-api/how-to-impersonate-any-address-as-a-simulation-sender

However I did not dig too much.

amusingaxl: Actually, never mind... impersonating won't solve the issue here. `lift` is permissionless, it requires sufficient approvals only.

We'd need to keep in mind that there's a new Chief implementation coming and the storage layout might change there, so this script would have to be updated.

scripts/cast-on-tenderly/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "deploy-to-tenderly",
+  "type": "module",
+  "version": "1.0.0",
+  "description": "A script to execute spell on a fork",
+  "main": "index.js",
+  "scripts": {
+    "do": "node ."
```
amusingaxl, line 8: Why not use the traditional `start` script here?

scripts/cast-on-tenderly/index.js
SidestreamColdMelon: To my understanding, because pause proxy have to do a delegate call, we can't impersonate pause proxy, we have to impersonate pause to call [pause-proxy's `exec`](https://github.com/dapphub/ds-pause/blob/0763eafcf926fd2e073aee5f047f3decb842231c/src/pause.sol#L114-L121). Then, [as `eta` is technically part of the spell state](https://github.com/makerdao/dss-exec-lib/blob/69b658f35d8618272cd139dfc18c5713caf6b96b/src/DssExec.sol#L41), it anyway have to be scheduled first. Maybe I'm also missing something, but impersonating pause proxy was also my first approach, which didn't worked out.

The other, classic approach would be to not use `setStorageAt` for the cheif's hat, but to mint enough mkr and then follow regular process of casting. Despite minting also requires `setStorageAt` under the hood, tenderly actually have more reliable abstraction of [`tenderly_setErc20Balance`](https://docs.tenderly.co/devnets/advanced/custom-rpc-methods#tenderly_seterc20balance). I actually would prefer this way, since then we can also use the same code to cast spell outside of the tenderly fork, but it will require more code 🙂

scripts/cast-on-tenderly/index.js
SidestreamColdMelon: I would be the same thing in the end (one hardcoded address on top of the file), but with more extra code to review. But we can do it, to make it a little bit less fragile and more robust to updates of the protocol

scripts/cast-on-tenderly/index.js
amusingaxl: Yeah, you are totally right. I'm gonna resolve this comment.
We just need to keep in mind that an update to the Chief might imply in maintenance for this script.

scripts/cast-on-tenderly/package.json
SidestreamColdMelon: changed to `start`!

Makefile
SidestreamColdMelon: I've updated it, thanks! FYI: in order to pass arguments to the underlying npm script, one have to use `--`, [see docs](https://docs.npmjs.com/cli/v8/commands/npm-run-script)

scripts/cast-on-tenderly/index.js
SidestreamColdMelon: Now only using hardcoding address of the chainlog

scripts/cast-on-tenderly/index.js
```diff
@@ -0,0 +1,110 @@
+import 'dotenv/config';
+import axios from 'axios';
+import { Contract, ethers } from 'ethers';
+
+const NETWORK_ID = 5;
+const CHAINLOG_ADDRESS = '0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F';
+const CHIEF_HAT_SLOT = 12;
+const DEFAULT_TRANSACTION_PARAMETERS = { gasLimit: 1000000000 };
```
amusingaxl, line 8: Just one small nit:

```suggestion
const DEFAULT_TRANSACTION_PARAMETERS = { gasLimit: 1_000_000_000 };
```

Node.js supports numeric separators [since v12.8.1](https://node.green/#ES2021-features--numeric-separators).

scripts/cast-on-tenderly/index.js
```diff
@@ -0,0 +1,110 @@
...
+    throw new Error(`Please provide all required env variables: ${REQUIRED_ENV_VARS.join(', ')}`);
+}
+
+// check process arguments
+const SPELL_ADDRESS = process.argv[2];
+if (!SPELL_ADDRESS) {
+    throw new Error('Please provide address of the spell, e.g.: `node index.js 0x...`');
+}
+
+const makeTenderlyApiRequest = async function (path) {
```
amusingaxl, line 22: Random question: why do you prefer this way for declaring a function instead of:

```js
async function makeTenderlyApiRequest(path) {}
```

or:

```js
const makeTenderlyApiRequest = async (path) => {}
```

SidestreamColdMelon: - To avoid unintentional use of hoisting (when function can be used before it is declared)
- `const` pattern also allows you to declare arrow functions the same way if you need to (because they are different from regular `function`s that arrow functions doesn't have `this` context)

amusingaxl (approved):
LGTM.

Good job @SidestreamColdMelon! 🚀

scripts/cast-on-tenderly/index.js
```diff
@@ -0,0 +1,110 @@
...
+const NETWORK_ID = 5;
+const CHAINLOG_ADDRESS = '0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F';
+const CHIEF_HAT_SLOT = 12;
+const DEFAULT_TRANSACTION_PARAMETERS = { gasLimit: 1_000_000_000 };
+
+// check env vars
+const REQUIRED_ENV_VARS = ['TENDERLY_USER', 'TENDERLY_PROJECT', 'TENDERLY_ACCESS_KEY'];
+if (REQUIRED_ENV_VARS.some(varName => !process.env[varName])) {
+    throw new Error(`Please provide all required env variables: ${REQUIRED_ENV_VARS.join(', ')}`);
+}
```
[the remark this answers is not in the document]
amusingaxl: Oh, looks like that module is not maintained anymore, someone [forked](https://www.npmjs.com/package/dotenv-safest) it and looks like they are [trying to make it part of the main `dotenv` package](https://github.com/motdotla/dotenv/issues/709).

README.md
```diff
@@ -52,3 +52,14 @@ $ export ETH_GAS_LIMIT=8000000
 $ export ETH_GAS_PRICE=$(seth --to-wei 3 "gwei")
 $ make deploy
 ```
+
+### Cast to tenderly
+
+1. Create Tenderly account (no trial period needed atm) https://dashboard.tenderly.co/register
+    - Note down `TENDERLY_USER` and `TENDERLY_PROJECT` values
```
SidestreamSweatyPumpkin, line 59: Nit: this reads like these values will only be accessible if you're creating the user. I would adjust this instruction to say something along the lines:

```
- get your username from https://dashboard.tenderly.co/account and note it down for later use as `TENDERLY_USER` env variable
- get your tenderly project name from https://dashboard.tenderly.co/projects and note it down to use later as `TENDERLY_PROJECT` env variable
```

SidestreamSweatyPumpkin (approved):
Just an optional nit. Otherwise lgtm :shipit:

README.md
SidestreamColdMelon: Thanks for providing extra instructions, but I would try to avoid writing a full tutorial on how to setup tenderly account, because we don't want to maintain it (and URLs and UIs tend to change over time)

scripts/cast-on-tenderly/index.js
```diff
@@ -0,0 +1,110 @@
...
+        await scheduleTx.wait();
+    } catch (error) {
+        console.warn('scheduling failed', error);
+    }
+
+    console.info('fetching timestamp when the spell will be castable...');
+    const eta = await spell.eta();
+
+    console.info(`warping the time to "${eta}"...`);
+    const currentUnixTimestamp = Date.now() / 1000;
```
SidestreamSweatyPumpkin, line 90: too late of a comment, but i think `parseInt` is due because it can result into float here and below we do `hexValue()` call on the division result

SidestreamColdMelon: Thanks for spotting it and noting it down here! 

Although I would suggest to use `Math.ceil()` over `parseInt()`. Because `parseInt()` is for [string parsing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt), not for rounding 🙂

SidestreamColdMelon: (this was later fixed by https://github.com/makerdao/spells-goerli/pull/229)

PR #227, opened 2023-08-31
Fix broken tests after casting a spell

amusingaxl:
## Context

The `setUp` function was supposed to take care of the scenario when the spell has already been cast using `rollFork`.

However, there seems to be an outstanding [issue](https://github.com/foundry-rs/foundry/issues/5739) on Foundry that prevents the proper setup.

The workaround is to manually overwrite the `eta` storage variable that is being read before using the `vm.store` cheatcode to ensure the spell can be scheduled and cast again.

---

Another small improvement: remove `vm.makePersistent(address(this))` from this function, because it is a superfluous call.

From the [docs](https://book.getfoundry.sh/cheatcodes/make-persistent?highlight=makeper#makepersistent):

> By default, only the test contract account and the caller are persistent across forks

### DIFF
--- a/src/DssSpell.t.base.sol
+++ b/src/DssSpell.t.base.sol
@@ -282,6 +282,7 @@ contract DssSpellTestBase is Config, DssTest {
 
     function setUp() public {
         setValues(address(chief));
+        _castPreviousSpell();
 
         spellValues.deployed_spell_created = spellValues.deployed_spell != address(0)
             ? spellValues.deployed_spell_created
@@ -289,8 +290,6 @@ contract DssSpellTestBase is Config, DssTest {
         spell = spellValues.deployed_spell != address(0)
             ?  DssSpell(spellValues.deployed_spell)
             : new DssSpell();
-
-        _castPreviousSpell();
 
         if (spellValues.deployed_spell_block != 0 && spell.eta() != 0) {
             // if we have a deployed spell in the config
--- a/src/DssSpell.t.base.sol
+++ b/src/DssSpell.t.base.sol
@@ -283,21 +283,33 @@ contract DssSpellTestBase is Config, DssTest {
     function setUp() public {
         setValues(address(chief));
 
-        spellValues.deployed_spell_created = spellValues.deployed_spell != address(0) ? spellValues.deployed_spell_created : block.timestamp;
+        spellValues.deployed_spell_created = spellValues.deployed_spell != address(0)
+            ? spellValues.deployed_spell_created
+            : block.timestamp;
+        spell = spellValues.deployed_spell != address(0)
+            ?  DssSpell(spellValues.deployed_spell)
+            : new DssSpell();
+
         _castPreviousSpell();
-        spell = spellValues.deployed_spell != address(0) ?
-            DssSpell(spellValues.deployed_spell) : new DssSpell();
 
         if (spellValues.deployed_spell_block != 0 && spell.eta() != 0) {
             // if we have a deployed spell in the config
             // we want to roll our fork to the block where it was deployed
             // this means the test suite will continue to accurately pass/fail
             // even if mainnet has already scheduled/cast the spell
-            vm.makePersistent(address(this));
             vm.makePersistent(address(rates));
             vm.makePersistent(address(addr));
             vm.makePersistent(address(deployers));
             vm.rollFork(spellValues.deployed_spell_block);
+
+            // Reset `eta` to `0`, otherwise the tests will fail with "This spell has already been scheduled".
+            // This is a workaround for the issue described here:
+            // @see { https://github.com/foundry-rs/foundry/issues/5739 }
+            vm.store(
+                address(spell),
+                bytes32(0),
+                bytes32(0)
+            );
         }
     }
 


0xdecr1pto (approved):
LGTM

The-Arbiter:
Questions:
- Formatting of ternary operators (OK, no question)
- Stop 'this' address from being persistent --> This is making it so that the state of the spell test contract (`this`? is not affected by block jump; or is this the spell contract?
- Reset spell's ETA to 0 (Allow it to be `scheduled` again)  --> Can we not just make the test a bit better by checking the state of the PAUSE (i.e. if a plan matching the hash of the spell we are working on has been plotted yet?)
This would break for nonconformant spells (see Kiln for conformant spec) where `eta` returned by the DssExec instance is false / a lie, which is roughly the same level of dishonesty as a 'tag' / extcodehash being falsified.

e.g. Can we check if the spell has been scheduled yet by querying `ds-pause` or is that issue also blocked by the foundry bug?

The-Arbiter:
Using some 'read' style fix as a workaround of a design flaw / bug in foundry is infinitely better than using a 'write' style fix (like vm.store) in general.

I think in this case we could probably say with 100% certainty that modifying 'eta' of the spell contract probably won't break anything, but in general, we should avoid modifying the state used in our integration tests _if_ there is a way to obtain the same result without modifying state (i.e. some read only way).

Note that the moment 'eta' returned by the spell isn't 'honest', the spell is not DssSpell conformant (i.e. spell address, extcodehash, calldata and ETA are all returned honestly from the spell), though after reading the discussions on [the foundry issue](https://github.com/foundry-rs/foundry/issues/5739) I agree with you that the expected behaviour of the helpers we are using is different to what is actually happening (this is probably why our tests were failing in the first place - we thought it worked one way and it worked another way).

I propose that we use the ETA of the spell plotted in the pause to determine if the spell has been scheduled or not.

If ETA in pause is > (the GSM delay) in the future, then the spell hasn't been plotted / scheduled yet
If ETA in pause is <= (the GSM delay) in the future, then the spell hasn't been plotted / scheduled yet

Not sure of the logistics required to make this work, but it's read only (doesn't use vm.store) and is a workaround.

FWIW I am not opposing the changes in principle, I just want us **not** to vm.store stuff if we don't absolutely have to

The-Arbiter:
Using some 'read' style fix as a workaround of a design flaw / bug in foundry is infinitely better than using a 'write' style fix (like vm.store) in general.

I think in this case we could probably say with 100% certainty that modifying 'eta' of the spell contract probably won't break anything, but in general, we should avoid modifying the state used in our integration tests _if_ there is a way to obtain the same result without modifying state (i.e. some read only way).

Note that the moment 'eta' returned by the spell isn't 'honest', the spell is not DssSpell conformant (i.e. spell address, extcodehash, calldata and ETA are all returned honestly from the spell), though after reading the discussions on [the foundry issue](https://github.com/foundry-rs/foundry/issues/5739) I agree with you that the expected behaviour of the helpers we are using is different to what is actually happening (this is probably why our tests were failing in the first place - we thought it worked one way and it worked another way).

I propose that we use the ETA of the spell plotted in the pause to determine if the spell has been scheduled or not.

If ETA in pause is > (the GSM delay) in the future, then the spell hasn't been plotted / scheduled yet
If ETA in pause is <= (the GSM delay) in the future, then the spell hasn't been plotted / scheduled yet

Not sure of the logistics required to make this work, but it's read only (doesn't use vm.store) and is a workaround.

FWIW I am not opposing the changes in principle, I just want us **not** to vm.store stuff if we don't absolutely have to

amusingaxl:
> Stop 'this' address from being persistent --> This is making it so that the state of the spell test contract (this? is not affected by block jump; or is this the spell contract?

No, because `address(this)` is **already** persistent across forks (see the docs and the highlighted excerpt), so we don't need to call it explicitly.

> Reset spell's ETA to 0 (Allow it to be scheduled again) --> Can we not just make the test a bit better by checking the state of the PAUSE (i.e. if a plan matching the hash of the spell we are working on has been plotted yet?)

I elaborate why I think this is not a good idea further down.

> This would break for nonconformant spells (see Kiln for conformant spec) where eta returned by the DssExec instance is false / a lie, which is roughly the same level of dishonesty as a 'tag' / extcodehash being falsified.

The tests in this repo are only supposed to work with conforming spells, what's the point of having safety checks against non-conforming ones? 

---

> Using some 'read' style fix as a workaround of a design flaw / bug in foundry is infinitely better than using a 'write' style fix (like vm.store) in general.
>
> I think in this case we could probably say with 100% certainty that modifying 'eta' of the spell contract probably won't break anything, but in general, we should avoid modifying the state used in our integration tests if there is a way to obtain the same result without modifying state (i.e. some read only way).

The entire point of that section of the `setUp` function is to revert the chain to the block where the spell was deployed. At that point in time, `eta` **IS** zero. That should be the state **AFTER** we call `rollFork()`.

Currently, this is not zero, only because Foundry is caching its value when it read in the conditional `if (spellValues.deployed_spell_block != 0 && spell.eta() != 0)`. If we remove the `spell.eta()` read from the check, it works as expected.

> FWIW I am not opposing the changes in principle, I just want us not to vm.store stuff if we don't absolutely have to

Another possibility would change `scheduleWaitAndCast` to not schedule if `eta` is not zero. Something like:

```solidity
    function _scheduleWaitAndCast(address spell_) internal {
        if (DssSpell(spell_).eta() == 0) {
            DssSpell(spell_).schedule();
        }

        vm.warp(DssSpell(spell_).nextCastTime());

        DssSpell(spell_).cast();
    }
```

This would fix the issue as well, however, the fix is now located somewhere else in the code, which hinders the locality principle: the issue is being generated by an allegedly faulty behavior in a call made from `setUp`, but is being fixed on `_scheduleWaitAndCast`.

This would be non-obvious for anyone that has not partaken into this original discussion, potentially adding it to the tech-debt pile down the road if the issue takes a long time to be fixed.

SidestreamSweatyPumpkin:
> The tests in this repo are only supposed to work with conforming spells

I would agree with this point. I do not see much purpose in covering the non-conforming ones with test since we cant expect any of the spec's standards to be applicable to them.

> This would fix the issue as well, however, the fix is now located somewhere else in the code, which hinders the locality principle

although true, we're patching something that does not work with a workaround here. Therefore, I don't have the strong opinion about which path / solution to choose here. Why the code is the way it is can be documented in the comment next to the code with the link to the foundry issue and this thread which allows to understand what is going on.

Please, arrive to a consensus regarding the implementation of the fix. If the second proposed option (code snippet in [this comment](https://github.com/makerdao/spells-goerli/pull/227#issuecomment-1705002247)) is chosen, please also add comments with contextual links.

With all above, the fix seems fine to me as it is now. I would like to merge it in before the next goerli spell review stage.

SidestreamSweatyPumpkin (approved):

amusingaxl:
There is yet another possibility, which is not checking `spell.eta()` at all.

If we change the conditional to:

```solidity
if (spellValues.deployed_spell_block != 0) {
    // ....
}
```

It would work as expected.

For the purpose of this setup, it really doesn't matter if the spell has been scheduled or not, since are rolling back the chain to the block where the spell was deployed. At that point, it's impossible for a spell to have been scheduled.

With the current condition, we only roll back if the scheduling has already happened, however there is no intermediate state that is relevant for the test suite. The only important piece is that we run the tests with the actual deployed spell, which we would do regardless.

With that change, we won't run into the Foundry issue and everything will work as expected.

PR #247, opened 2024-02-06
Fix: cast spell script

amusingaxl:
The script is broken since Foundry introduced an additional output when specifying the return type of the function in the signature:

```bash
cast call "$MCD_GOV" 'balanceOf(address)(uint)' "$ETH_FROM"
```

Now returns:

```
100000000000000000000000 [1e23]
```

Notice the dangling ` [1e23]`.

Previously, it would return simply:

```
100000000000000000000000
```

I'm not entirely sure when this change was introduced, but it must be a recent one, since last time I cast a Goerli spell (late 2023), it worked without any issues.


amusingaxl:
A bit more context:
https://github.com/foundry-rs/foundry/issues/6600
https://github.com/foundry-rs/foundry/issues/6982
https://github.com/foundry-rs/foundry/issues/6598

SidestreamColdMelon (approved):

oddaf (approved):

